PR #22, opened 2024-10-09
feat(iOS): scroll edge appearance

dylancom:
In many iOS apps such as the health app, the tabbar becomes transparent (without border).
In this PR I am looking into how we can add this.
It currently seems to bug at first render (as the comment also says: _This causes issues with lazy loading making the TabView background blink._), but if you scroll once the effect comes in.

| **Transparent without content overlapping** | **Content overlap** |
|---------------------------------------------|---------------------|
| Simulator Screenshot - iPhone 16 Pro - 2024-10-09 at 08 24 45 | Simulator Screenshot - iPhone 16 Pro - 2024-10-09 at 08 25 37 |


okwasniewski:
Hey! Thanks for opening the PR.

I definitely want to support this but with lazy loading the tab bar was flashing before the content rendered.

I think we could make this into a prop, something like: `showTabBarBackground` (or something similar) and by default manipulate this setting on the JS side to prevent flashing if user is using lazy loading.

WDYT?

dylancom:
@okwasniewski I am a bit unsure how to call this. Should we have a prop mimic the native side? `scrollEdgeAppearance="default | transparent"`.

_scrollEdgeAppearance: Describes the appearance attributes for the tabBar to use when an observable scroll view is scrolled to the bottom. If not set, standardAppearance will be used instead._

okwasniewski:
@dylancom To me scrollEdgeAppearance="default | transparent" makes sense. This will be iOS only so using native name might be easier

src/TabView.tsx
```diff
@@ -36,6 +36,10 @@ interface Props<Route extends BaseRoute> {
    * Whether to disable page animations between tabs. (iOS only)
    */
   disablePageAnimations?: boolean;
+  /**
+   * Describes the appearance attributes for the tabBar to use when an observable scroll view is scrolled to the bottom. (iOS only)
+   */
```
okwasniewski, line 41: Can you also add this to the docs? 

https://github.com/okwasniewski/react-native-bottom-tabs/blob/main/docs/docs/docs/guides/usage-with-react-navigation.md

ios/TabViewProvider.swift
```diff
@@ -59,6 +59,12 @@ struct TabData: Codable {
     }
   }
   
+@objc var scrollEdgeAppearance: NSString? {
+    didSet {
+      props.scrollEdgeAppearance = scrollEdgeAppearance as? String
+    }
+  }
```
okwasniewski, line 66: Something is off with formatting here

ios/TabViewImpl.swift
```diff
@@ -73,8 +74,16 @@ struct TabViewImpl: View {
...
+        let scrollEdgeAppearance = UITabBarAppearance()
+        switch props.scrollEdgeAppearance {
+        case "opaque":
+          scrollEdgeAppearance.configureWithOpaqueBackground()
+        case "transparent":
+          scrollEdgeAppearance.configureWithTransparentBackground()
+        default:
+          scrollEdgeAppearance.configureWithDefaultBackground()
+        }
+        UITabBar.appearance().scrollEdgeAppearance = scrollEdgeAppearance
```
okwasniewski, line 86: Can you move this to a helper function?

ios/RCTTabViewViewManager.mm
```diff
@@ -31,5 +31,6 @@ - (UIView *)view
 RCT_EXPORT_VIEW_PROPERTY(labeled, BOOL)
 RCT_EXPORT_VIEW_PROPERTY(ignoresTopSafeArea, BOOL)
 RCT_EXPORT_VIEW_PROPERTY(disablePageAnimations, BOOL)
+RCT_EXPORT_VIEW_PROPERTY(scrollEdgeAppearance, NSString)
```
okwasniewski, line 34: You also need to add it to JS Spec here: https://github.com/okwasniewski/react-native-bottom-tabs/blob/95782ac53c09dd5d42b2ece0aae877310c762857/src/TabViewNativeComponent.ts#L18

okwasniewski (commented):
Thanks for working on this! Looks good, I left few comments

dylancom:
Created a new clean PR, see: #54

PR #30, opened 2024-10-09
[wip] feat: handle bottom safe area

okwasniewski:


dylancom:
I am experimenting now with switching to Swift / UIKit and noticed that scrollviews respect the TabBar / Safearea. I think this is more viable than hardcoding values?
See: #31

okwasniewski:
Yeah, hardcoding is not ideal maybe we could measure the tab bar.

According to Apple and Stack overflow it's constant at those values so this option might work. I couldn't find any working native solution yet

dylancom:
I'm not sure if this helps in the right direction, but I was now able to adjust the height of "Article" from within SwiftUI.
Forcing a height of 600 (which we could adjust to the height minus tabbar) and the scrollview adjusts to it.
So we might have to set some constraints?

```
struct RepresentableView: UIViewRepresentable {
  var view: UIView
  
  func makeUIView(context: Context) -> UIView {
    applyConstraintsToSubviews(of: view)
    if let subview = view.subviews.first {
      applyConstraintsToSubviews(of: subview)

      if let subview2 = subview.subviews.first {
        applyConstraintsToSubviews(of: subview2)
      }
    }
    return view
  }
  func updateUIView(_ uiView: UIView, context: Context) {
  }
  
  private func applyConstraintsToSubviews(of parentView: UIView) {
    parentView.translatesAutoresizingMaskIntoConstraints = false

    NSLayoutConstraint.activate([
      parentView.widthAnchor.constraint(equalToConstant: parentView.superview?.bounds.width ?? 0),
      parentView.heightAnchor.constraint(equalToConstant: 600) // Force height limit
    ])
  }
}
```

PR #31, opened 2024-10-10
fix: switch from SwiftUI to Swift, fixes safe area handling

dylancom:
It seems that switching from SwiftUI to Swift/UIKit, the ScrollView respects the TabBar...

### DIFF
--- a/ios/TabViewImpl.swift
+++ b/ios/TabViewImpl.swift
@@ -1,114 +1,69 @@
-import Foundation
-import SwiftUI
-import React
+import UIKit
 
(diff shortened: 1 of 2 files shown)


okwasniewski:
Hey, thanks for opening the PR but switching to UIKit is not an viable option as this will block macOS support + sidebarAdaptable is not available for UIKit.

okwasniewski:
> Hmm atleast this shows there is a better way than hardcoding TabBar heights...

Yeah definitely, let's try to find a better way

PR #37, opened 2024-10-11
Fix/bottom safe area

okwasniewski:
This PR fixes bottom safe area handling by measuring the tab bar on the native side and offsetting it in JS.

https://github.com/user-attachments/assets/5973256f-26b9-46aa-94a8-2c420a468517

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -1209,7 +1209,7 @@ PODS:
     - ReactCommon/turbomodule/bridging
     - ReactCommon/turbomodule/core
     - Yoga
-  - react-native-bottom-tabs (0.1.0):
+  - react-native-bottom-tabs (0.0.9):
     - DoubleConversion
     - glog
     - RCT-Folly (= 2024.01.01.00)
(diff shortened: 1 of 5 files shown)


okwasniewski:
cc @dylancom let me know what you think 👍🏻

dylancom:
Will look into more details later. Does this work with "overflow: visible" on a scrollview/flatlist? In the video it seems that the content isn't blurred below the tabbar.

dylancom:
In order to make the blurring backgorund on the tabbar work `overflow: visible` needs to be set on the scrollviews.
A bit a shame that this is working out of the box when switching to UIKit.

Simulator Screenshot - iPhone 16 Pro - 2024-10-11 at 15 51 05

TabMeasuredEvent.swift
dylancom, line 1: File should be moved to ios/ folder. Currently getting an error that TabMeasuredEvent can't be found.

dylancom (changes requested):

okwasniewski:
Hey @dylancom I finally figured it out!! 

Now its working with the built-in behavior 🚀


https://github.com/user-attachments/assets/76082f3a-e0c9-4e2e-b8f9-4f14f395c099

dylancom:
Clever find!!
Awesome.

dylancom:
Some bugs while testing on my end. The scrollbar starts too late in the top and on reaching the end some content is still below the tabbar.

dylancom:
If I test the example app in the latest main the content still ends below the tabbar on my end.

ferrannp:
@okwasniewski I just tried this and the issue still persists. Do you have a working example?

jnsdrssn:
@ferrannp try `contentInsetAdjustmentBehavior="always"` on the `ScrollView`

PR #42, opened 2024-10-14
Adding Active Inactive Tint Color

Stringsaeed:
### PR Description:
This pull request introduces a new feature that adds support for active/inactive tint colors on tabs across both Android and iOS components.

### Notes:
- **Color Handling:**  
  The colors are sourced from the `react-navigation` theme by default. This follows the same approach used by `@react-navigation/bottom-tabs`. As a result, I added dependencies for `color` and `@types/color` to handle this functionality.

- **Android Implementation:**  
  My Kotlin code might feel more Java-like, as I'm more familiar with Java. So, you’ll notice some usage of `as` and `if (x != null)` for type narrowing.

- **iOS Implementation:**  
  As you may know, SwiftUI Tabs rely on the appearance of the UIKit tab bar. However, SwiftUI doesn't provide a built-in way to customize the inactive (unselected) tint color. To work around this, I had to leverage the UIKit appearance settings.

### DIFF
--- a/android/src/main/java/com/rcttabview/RCTTabView.kt
+++ b/android/src/main/java/com/rcttabview/RCTTabView.kt
@@ -32,6 +32,10 @@ class ReactBottomNavigationView(context: Context) : BottomNavigationView(context
   private var isAnimating = false
   private var activeTintColor: Int? = null
   private var inactiveTintColor: Int? = null
+  private val checkedStateSet = intArrayOf(android.R.attr.state_checked)
+  private val uncheckedStateSet = intArrayOf(-android.R.attr.state_checked)
+  private val disabledStateSet = intArrayOf(-android.R.attr.state_enabled)
+
 
   private val layoutCallback = Choreographer.FrameCallback {
     isLayoutEnqueued = false
@@ -162,32 +166,6 @@ class ReactBottomNavigationView(context: Context) : BottomNavigationView(context
     itemBackground = colorDrawable
   }
 
-  private fun getDefaultColorFor(baseColorThemeAttr: Int): Int? {
-    val value = TypedValue()
-    if (!context.theme.resolveAttribute(baseColorThemeAttr, value, true)) {
-      return null
-    }
-    val baseColor = AppCompatResources.getColorStateList(context, value.resourceId)
-    return baseColor.defaultColor
-  }
-
-  private fun updateTintColors() {
-    val activeColor = activeTintColor ?: return
-    val inactiveColor = inactiveTintColor ?: return
-
-    val states = arrayOf(
-      intArrayOf(-android.R.attr.state_checked),
-      intArrayOf(android.R.attr.state_checked)
-    )
-
-    val colors = intArrayOf(inactiveColor, activeColor)
-
-    ColorStateList(states, colors).apply {
-      this@ReactBottomNavigationView.itemTextColor = this
-      this@ReactBottomNavigationView.itemIconTintList = this
-    }
-  }
-
   fun setActiveTintColor(color: Int?) {
     activeTintColor = color
     updateTintColors()
@@ -196,5 +174,30 @@ class ReactBottomNavigationView(context: Context) : BottomNavigationView(context
   fun setInactiveTintColor(color: Int?) {
     inactiveTintColor = color
     updateTintColors()
+  }
+
+  private fun updateTintColors() {
+    // getDeaultColor will always return a valid color but to satisfy the compiler we need to check for null
+    val colorPrimary = activeTintColor ?: getDefaultColorFor(android.R.attr.colorPrimary) ?: return
+    val colorSecondary =
+      inactiveTintColor ?: getDefaultColorFor(android.R.attr.textColorSecondary) ?: return
+    val states = arrayOf(uncheckedStateSet, checkedStateSet)
+    val colors = intArrayOf(colorSecondary, colorPrimary)
+
+    ColorStateList(states, colors).apply {
+      this@ReactBottomNavigationView.itemTextColor = this
+      this@ReactBottomNavigationView.itemIconTintList = this
+    }
+  }
+
+  private fun getDefaultColorFor(baseColorThemeAttr: Int): Int? {
+    val value = TypedValue()
+    if (!context.theme.resolveAttribute(baseColorThemeAttr, value, true)) {
+      return null
+    }
+    val baseColor = AppCompatResources.getColorStateList(
+      context, value.resourceId
+    )
+    return baseColor.defaultColor
   }
 }
--- a/android/src/main/java/com/rcttabview/RCTTabView.kt
+++ b/android/src/main/java/com/rcttabview/RCTTabView.kt
@@ -1,6 +1,5 @@
 package com.rcttabview
 
-import android.R.attr
 import android.content.Context
 import android.content.res.ColorStateList
 import android.graphics.Color
@@ -173,33 +172,29 @@ class ReactBottomNavigationView(context: Context) : BottomNavigationView(context
   }
 
   private fun updateTintColors() {
-    if (activeTintColor == null || inactiveTintColor == null) {
-      return;
-    }
+    val activeColor = activeTintColor ?: return
+    val inactiveColor = inactiveTintColor ?: return
 
     val states = arrayOf(
-      intArrayOf(-attr.state_checked),
-      intArrayOf(attr.state_checked)
-    )
-    val colors = intArrayOf(
-      inactiveTintColor as Int,
-      activeTintColor as Int
+      intArrayOf(-android.R.attr.state_checked),
+      intArrayOf(android.R.attr.state_checked)
     )
 
-    this.itemTextColor = ColorStateList(states, colors)
+    val colors = intArrayOf(inactiveColor, activeColor)
 
-    this.itemIconTintList = ColorStateList(
-      states, colors
-    )
+    ColorStateList(states, colors).apply {
+      this@ReactBottomNavigationView.itemTextColor = this
+      this@ReactBottomNavigationView.itemIconTintList = this
+    }
   }
 
   fun setActiveTintColor(color: Int?) {
-      activeTintColor = color
-      this.updateTintColors()
+    activeTintColor = color
+    updateTintColors()
   }
 
   fun setInactiveTintColor(color: Int?) {
-      inactiveTintColor = color
-      this.updateTintColors()
+    inactiveTintColor = color
+    updateTintColors()
   }
 }
--- a/android/src/main/java/com/rcttabview/RCTTabView.kt
+++ b/android/src/main/java/com/rcttabview/RCTTabView.kt
@@ -1,5 +1,6 @@
 package com.rcttabview
 
+import android.R.attr
 import android.content.Context
 import android.content.res.ColorStateList
 import android.graphics.Color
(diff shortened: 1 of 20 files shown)


okwasniewski:
Thanks for working on this! 

I'll try to review this PR in the upcoming days 🙌

okwasniewski:
For some reasons examples other than "Tint Colors" don't work for me. Can you do a rebase and check if that's the case for you?

Stringsaeed:
@okwasniewski fixed ✅

android/src/main/java/com/rcttabview/RCTTabView.kt
```diff
@@ -140,4 +142,31 @@ class ReactBottomNavigationView(context: Context) : BottomNavigationView(context
     super.onDetachedFromWindow()
     isAnimating = false
   }
+
+  private fun updateTintColors() {
+    val activeColor = activeTintColor ?: return
+    val inactiveColor = inactiveTintColor ?: return
```
okwasniewski, line 148: What if someone only sets activeColor or inactiveColor? In that case this won't work.

ios/TabViewImpl.swift
```diff
@@ -59,15 +61,29 @@ struct TabViewImpl: View {
...
+        // @see https://stackoverflow.com/a/71934882
+        if let inactiveTintColor = props.inactiveTintColor {
+          appearance.stackedLayoutAppearance.normal.titleTextAttributes = [.foregroundColor: inactiveTintColor]
+          appearance.stackedLayoutAppearance.normal.iconColor = inactiveTintColor
+        }
+        
+        if let activeTintColor = props.activeTintColor {
+          appearance.stackedLayoutAppearance.selected.titleTextAttributes = [.foregroundColor: activeTintColor]
+          appearance.stackedLayoutAppearance.selected.iconColor = activeTintColor
+        }
```
okwasniewski, line 82: Isn't activeTintColor handled by `tintColor` modifider?

android/src/main/java/com/rcttabview/RCTTabView.kt
Stringsaeed: Either that or we have to set some default value of both colors as we can't make `ColorStateList` with null values

Maybe we should access material theme value and add it as default value WDYT?

ios/TabViewImpl.swift
Stringsaeed: Updated ✅

android/src/main/java/com/rcttabview/RCTTabView.kt
Stringsaeed: Updated ✅

okwasniewski:
Thanks for bringing it the fixes! 

I looked into JS Bottom Tabs and it looks like they have `tabBarActiveTintColor` and `tabBarInactiveTintColor` as options that you can pass to the screen. So you can technically have different tints for each item. I would prefer to match the JS bottom tabs API as closely as possible. Do you think we can change the way it works to be per tab?

android/src/main/java/com/rcttabview/RCTTabView.kt
okwasniewski: Looks good!

Stringsaeed:
Yes we can add the same logic per item on select event but I don't know if it will be the best implementation for the performance 

I will experiment it and push the code for it

example/src/Examples/NativeBottomTabs.tsx
```diff
@@ -9,7 +9,7 @@ const Tab = createNativeBottomTabNavigator();
 
 function NativeBottomTabs() {
   return (
-    <Tab.Navigator>
+    <Tab.Navigator inactiveTintColor="blue">
```
shubhamguptadream11, line 12: For JS Bottom tabs we have 2 props: `tabBarActiveTintColor`, `tabBarInActiveTintColor` to provide this functionality, which are part of screenOptions.

To make API close to JS one don't you think we should include this inside screenOptions and match names as well?

Stringsaeed:
@okwasniewski 

The thing is in SwiftUI we can't have different color for each tab, we will have to make it only different tint when selected only but we can't have different inactive color as there's no  a modifier to say that color for inactive color

So when we define inactive color in each screen we will have to use that for our current all other tab active color that's the current implementation draft I have:

```swift

func updateUITabBarAppearance() {
  let appearance = UITabBarAppearance()

  if let inactiveTintColor = getInactiveTintColor()  {
    appearance.stackedLayoutAppearance.normal.titleTextAttributes = [.foregroundColor: inactiveTintColor]
    appearance.stackedLayoutAppearance.normal.iconColor = inactiveTintColor
  }

  if let activeTintColor = getActiveTintColor() {
    appearance.stackedLayoutAppearance.selected.titleTextAttributes = [.foregroundColor: activeTintColor]
  }
    
  if #available(iOS 15.0, *) {
    // This causes issues with lazy loading making the TabView background blink.
    UITabBar.appearance().scrollEdgeAppearance = appearance
  }

  UITabBar.appearance().standardAppearance = appearance
}
```

and this will result in different effect than JS one for SwiftUI it's a limitation to just follow [Apple Design Guide](https://developer.apple.com/design/human-interface-guidelines/tab-bars) 

cc: @shubhamguptadream11

okwasniewski:
@Stringsaeed 

So activeTintColor can be handled using SwiftUI only, there is no need to additional UIKit code as we can use `.tintColor()` 

We can also easily achieve unselectedTintColor but for all tabs: 

`UITabBar.appearance().unselectedItemTintColor = inactiveTintColor`

Maybe we can make activeTintColor per tab item and inactive for the whole tab bar? Im not sure if there is any better way as it looks like a limitation of UITabBar

Stringsaeed:
@okwasniewski 


Unfortunately we can't use this
```swift
UITabBar.appearance().unselectedItemTintColor = inactiveTintColor
```
As we define our own `appearance`

**Here is an example of:**

| no override for `appearance`  | override for `appearance` |
|--------|--------|
| Screenshot 2024-10-17 at 11 09 38 PM | Screenshot 2024-10-17 at 11 09 27 PM |


So we have to stick with our method for inactive tint colors

src/react-navigation/navigators/createNativeBottomTabNavigator.tsx
```diff
@@ -34,8 +36,21 @@ function NativeBottomTabNavigator({
   children,
   screenListeners,
   screenOptions,
+  activeTintColor: customActiveTintColor,
+  inactiveTintColor: customInactiveTintColor,
```
okwasniewski, line 40: Can we match prop names from JS tabs? 

`tabBarActiveTintColor`
Color for the icon and label in the active tab.

`tabBarInactiveTintColor`
Color for the icon and label in the inactive tabs.

Also can you add those props to docs? 

https://github.com/okwasniewski/react-native-bottom-tabs/blob/main/docs/docs/docs/guides/usage-with-react-navigation.md

src/TabViewNativeComponent.ts
```diff
@@ -23,6 +23,16 @@ export interface TabViewProps extends ViewProps {
...
 }
 
+export type TabViewNativeProps = Omit<
+  TabViewProps,
+  'activeTintColor' | 'inactiveTintColor'
+> & {
+  activeTintColor?: ColorValue;
+  inactiveTintColor?: ColorValue;
+};
+
```
okwasniewski, line 37: I'm afraid this may fail during codegen with new architecture. 

Can you try installing pods with RCT_NEW_ARCH_ENABLED=1 to check if codegen generates specs correctly?

ios/TabViewImpl.swift
```diff
@@ -90,7 +94,17 @@ private func configureAppearance(for appearanceType: String) -> UITabBarAppearan
   default:
     appearance.configureWithDefaultBackground()
   }
-  
+
+  // @see https://stackoverflow.com/a/71934882
+  if let inactiveTintColor = props.inactiveTintColor {
+    appearance.stackedLayoutAppearance.normal.titleTextAttributes = [.foregroundColor: UIColor(inactiveTintColor)]
```
okwasniewski, line 100: You are applying this to `stackedLayoutAppearance` I think we need to apply it to every appearance. This might stop working when phone is rotated.

Reference: https://emptytheory.com/2019/12/31/using-uitabbarappearance-for-tab-bar-changes-in-ios-13/

ios/TabViewImpl.swift
```diff
@@ -90,7 +94,17 @@ private func configureAppearance(for appearanceType: String) -> UITabBarAppearan
...
+    appearance.stackedLayoutAppearance.normal.iconColor = UIColor(inactiveTintColor)
+  }
+  if let activeTintColor = props.activeTintColor {
+    appearance.stackedLayoutAppearance.selected.titleTextAttributes = [.foregroundColor: UIColor(activeTintColor)]
+    appearance.stackedLayoutAppearance.selected.iconColor = UIColor(activeTintColor)
```
okwasniewski, line 105: Is this required if we have `tintColor` modifier?

ios/TabViewImpl.swift
```diff
@@ -75,13 +78,14 @@ struct TabViewImpl: View {
       if #available(iOS 15.0, *) {
         UITabBar.appearance().scrollEdgeAppearance = configureAppearance(for: newValue ?? "")
       }
+      UITabBar.appearance().standardAppearance = configureAppearance(for: newValue ?? "")
```
okwasniewski, line 81: It would be best if you created a separate onChange() because this reacts only to `props.scrollEdgeAppearance`

It would be best if you had separate onChange for each prop

ios/RCTTabViewViewManager.mm
```diff
@@ -32,5 +32,9 @@ - (UIView *)view
 RCT_EXPORT_VIEW_PROPERTY(ignoresTopSafeArea, BOOL)
 RCT_EXPORT_VIEW_PROPERTY(disablePageAnimations, BOOL)
 RCT_EXPORT_VIEW_PROPERTY(scrollEdgeAppearance, NSString)
+RCT_EXPORT_VIEW_PROPERTY(activeTintColor, NSNumber)
+RCT_EXPORT_VIEW_PROPERTY(inactiveTintColor, NSNumber)
+RCT_EXPORT_VIEW_PROPERTY(activeTintColor, NSNumber)
+RCT_EXPORT_VIEW_PROPERTY(inactiveTintColor, NSNumber)
```
okwasniewski, line 38: Duplicated properties

okwasniewski:
So as I mentioned before, we can do: 

```tsx
function NativeBottomTabs() {
  return (
    <Tabs.Navigator tabBarInactiveTintColor="orange">
      <Tabs.Screen
        name="index"
        options={{
          title: "Home",
          tabBarIcon: () => ({ sfSymbol: "house" }),
          tabBarActiveTintColor: "red"
        }}
      />
      <Tabs.Screen
        name="explore"
        options={{
          title: "Explore",
          tabBarIcon: () => ({ sfSymbol: "person" }),
          tabBarActiveTintColor: "orange"
        }}
      />
    </Tabs.Navigator>
  );
}
```

Separate tint color for each tab, but one inactive color. This is a tradeoff but we have to stick to what native components allow us to achieve. (Let's describe this decision in docs)

okwasniewski:
Hey @Stringsaeed, 

I don't want to rush you in any way but this is a crucial feature for the library and users are asking me for it. 

Let me know if you currently don't have the time to work on this so someone else can pick it up where you left off. (cc: @shubhamguptadream11 @sarthak-d11)

Stringsaeed:
@okwasniewski 
Actually I was going to finish it today you can except an update in 10 hours around 9:00 PM (Poland time)

okwasniewski:
> @okwasniewski Actually I was going to finish it today you can except an update in 10 hours around 9:00 PM (Poland time)

Awesome, thank you @Stringsaeed 🎉

okwasniewski:
Thanks for working on this!

I've updated the PR with docs changes and fixes for both platforms. 

Now we get support for: 

- activeTintColor (navigator level)
- inactiveTintColor (navigator level)
- activeTintColor (per tab) 


https://github.com/user-attachments/assets/787dbc76-ea69-4a6c-a5db-ab42dd6a9c17

okwasniewski (approved):
Thank you!

PR #43, opened 2024-10-14
fix: Add ScrollView adjustment automatic

okwasniewski:
This PR fixes content insets in examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,16 @@ export default function ThreeTabs() {
 }
 ```
 
+### ScrollView
+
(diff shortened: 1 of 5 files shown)


okwasniewski:
cc: @dylancom

Forgot to add this, but this fixes the issue

dylancom:
Perfect!

dylancom:
@okwasniewski Still the scrollbar is offset approx 30px. The content isn't offset but the bar on the right for some reason.
AFAIK it should touch the top?

okwasniewski:
@dylancom My guess is that its the issue with how Native Stack lays out screens and works with safe area if you ignore safe area insets then it goes below the top bar. Probably this needs a separate issue

PR #44, opened 2024-10-14
fix: add react-navigation export to package.json

natew:
Yo! Awesome work with this library, was testing it out with onestack.dev and this makes things a bit easier to set up with Vite.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,16 @@
         "types": "./lib/typescript/commonjs/src/index.d.ts",
         "default": "./lib/commonjs/index.js"
       }
+    },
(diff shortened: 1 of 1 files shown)


natew:
I’m not sure I think that may be for metro or older node - what I’ve typically done is just added a new file like react-native.js at the root that they will use and that file just re-exports things from the right dist folder. I can link to that pattern later when I’m back from dinner.

okwasniewski (approved):
Thanks for the PR, let's leave the react-navigation/package.json to keep compatibility with older node as you mentioned

PR #52, opened 2024-10-16
fix: docs link target in header

jbroma:
fixed a link in the header based on config in Re.Pack [here](https://github.com/callstack/repack/blob/main/website/src/4.x/_meta.json)


okwasniewski (approved):
Thank you!

PR #54, opened 2024-10-17
feat(iOS): scroll edge appearance

dylancom:
This let's you define the edge appearance of the tabbar.
Allowing for a tabbar that is transparent when there is no content behind the tabbar.

@okwasniewski I am a bit in doubt what we should initially set this to... It seems that if you have a tabbarcontroller it modifies this to transparent by default? Do we want it transparent by default (on the scroll edge) or not... hmm...

_If the value of this property is nil, UIKit uses the value of the tab bar’s [standardAppearance](https://developer.apple.com/documentation/uikit/uitabbar/3198046-standardappearance) property, modified to have a transparent background. If no tab bar controller manages your tab bar, UIKit ignores this property and uses the tab bar’s standard appearance._

### DIFF
--- a/docs/docs/docs/guides/usage-with-react-navigation.md
+++ b/docs/docs/docs/guides/usage-with-react-navigation.md
@@ -55,6 +55,15 @@ Whether to show labels in tabs. Defaults to true.
 
 Whether to disable page animations between tabs. (iOS only)
 
+#### `scrollEdgeAppearance`
+
+Describes the appearance attributes for the tabBar to use when an observable scroll view is scrolled to the bottom. (iOS only)
+
+Available options:
+- `default` - uses default background and shadow values.
+- `transparent` - uses transparent background and no shadow.
+- `opaque` - uses set of opaque colors that are appropriate for the current theme
+
 #### `sidebarAdaptable`
 
 A tab bar style that adapts to each platform. (Apple platforms only)
--- a/docs/docs/docs/guides/usage-with-react-navigation.md
+++ b/docs/docs/docs/guides/usage-with-react-navigation.md
@@ -64,6 +64,7 @@ Available options:
 - `transparent` - uses transparent background and no shadow.
 - `opaque` - uses set of opaque colors that are appropriate for the current theme
 
+Note: It's recommended to use `transparent` or `opaque` without lazy loading as the tab bar background flashes when a view is rendered lazily.
 #### `sidebarAdaptable`
 
 A tab bar style that adapts to each platform. (Apple platforms only)
--- a/ios/TabViewImpl.swift
+++ b/ios/TabViewImpl.swift
@@ -14,6 +14,7 @@ class TabViewProps: ObservableObject {
   @Published var labeled: Bool?
   @Published var ignoresTopSafeArea: Bool?
   @Published var disablePageAnimations: Bool = false
+  @Published var scrollEdgeAppearance: String?
 }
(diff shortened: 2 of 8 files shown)


okwasniewski:
Hmm I don't really see the difference between opaque and transparent. Can we make it a booean? something like: `transparentScrollEdgeApperance`? 

I think setting (`configureWithDefaultBackground()`) as default makes sense

ios/TabViewImpl.swift
```diff
@@ -73,13 +74,27 @@ struct TabViewImpl: View {
     .onAppear {
```
okwasniewski, line 74: You should change it to `.onChange(of: props.scrollEdgeAppearance ?? "") {}` to allow for dynamic changing currently users can set this value only on first mount.

dylancom:
Haven't tested opaque clearly but thought it was neat to expose all available options as it was a small effort.

dylancom:
Is the lazy loading at all needed with this library (perf wise)? As in using it as default...

okwasniewski:
> Is the lazy loading at all needed with this library (perf wise)? As in using it as default...

It is, if someone renders a huge list in one of the tabs then it can cause issues. Also lazy is the default on JS Tabs so we are matching the "standard".

Maybe Im missing why there is transparent and opaque so let's expose it as you mentioned.

dylancom:
Yes, default is currently our default :)

docs/docs/docs/guides/usage-with-react-navigation.md
```diff
@@ -55,6 +55,10 @@ Whether to show labels in tabs. Defaults to true.
 
 Whether to disable page animations between tabs. (iOS only)
 
+#### `scrollEdgeAppearance`
+
+Describes the appearance attributes for the tabBar to use when an observable scroll view is scrolled to the bottom. (iOS only)
```
okwasniewski, line 60:
```suggestion
Describes the appearance attributes for the tabBar to use when an observable scroll view is scrolled to the bottom. (iOS only)

Available options:
- `default` - uses default background and shadow values.
- `transparent` - uses transparent background and no shadow.
- `opaque` - uses set of opaque colors that are appropriate for the current theme
```

ios/TabViewImpl.swift
```diff
@@ -70,16 +71,30 @@ struct TabViewImpl: View {
       }
       onSelect(newValue)
     }
-    .onAppear {
+    .onChange(of: props.scrollEdgeAppearance) { newValue in
       if #available(iOS 15.0, *) {
         // This causes issues with lazy loading making the TabView background blink.
```
okwasniewski, line 76:
```suggestion
```

docs/docs/docs/guides/usage-with-react-navigation.md
```diff
@@ -55,6 +55,15 @@ Whether to show labels in tabs. Defaults to true.
...
+
+Available options:
+- `default` - uses default background and shadow values.
+- `transparent` - uses transparent background and no shadow.
+- `opaque` - uses set of opaque colors that are appropriate for the current theme
+
```
okwasniewski, line 66:
```suggestion

Note: It's recommended to use `transparent` or `opaque` without lazy loading as the tab bar background flashes when a view is rendered lazily.
```

okwasniewski (approved):
Thanks for working on this! Two last docs changes

PR #55, opened 2024-10-18
Added support for barTintColor in android ios

shubhamguptadream11:
Part of this: https://github.com/okwasniewski/react-native-bottom-tabs/issues/11

**Summary:** This PR updates the tab bar background color for both iOS and Android implementations.

**Changes Made:**

**iOS**: Added the updateTabBarAppearance(with:) function to set the background color using UITabBarAppearance for iOS 15 and barTintColor for earlier versions

**Android**: Updated the tab bar background color to apply uniformly.

**Testing:**

- Verified background color changes on both iOS and Android.
- Added `FourTabsWithBarTintColor` for visual changes.

Adding snapshot of changes here:

iOS:
simulator_screenshot_6FD154D1-C2C9-4DD2-B6E9-50B27E680B62

Android:
Screenshot_20241018_142226

### DIFF
--- a/android/src/main/java/com/rcttabview/RCTTabView.kt
+++ b/android/src/main/java/com/rcttabview/RCTTabView.kt
@@ -145,10 +145,6 @@ class ReactBottomNavigationView(context: Context) : BottomNavigationView(context
   }
 
   fun setBarTintColor(color: Int?) {
-    updateBackgroundColors(color)
-  }
-
-  private fun updateBackgroundColors(color: Int?) {
     // Set the color, either using the active background color or a default color.
     val backgroundColor = color ?: getDefaultColorFor(android.R.attr.colorPrimary) ?: return
 
@@ -157,7 +153,6 @@ class ReactBottomNavigationView(context: Context) : BottomNavigationView(context
 
     itemBackground = colorDrawable
   }
-
 
   private fun getDefaultColorFor(baseColorThemeAttr: Int): Int? {
     val value = TypedValue()
--- a/android/src/main/java/com/rcttabview/RCTTabView.kt
+++ b/android/src/main/java/com/rcttabview/RCTTabView.kt
@@ -28,7 +28,6 @@ class ReactBottomNavigationView(context: Context) : BottomNavigationView(context
   var items: MutableList<TabInfo>? = null
   var onTabSelectedListener: ((WritableMap) -> Unit)? = null
   private var isAnimating = false
-  private var barTintColor : Int? = null
 
   private val layoutCallback = Choreographer.FrameCallback {
     isLayoutEnqueued = false
@@ -146,13 +145,12 @@ class ReactBottomNavigationView(context: Context) : BottomNavigationView(context
   }
 
   fun setBarTintColor(color: Int?) {
-    barTintColor = color
-    updateBackgroundColors()
+    updateBackgroundColors(color)
   }
 
-  private fun updateBackgroundColors() {
+  private fun updateBackgroundColors(color: Int?) {
     // Set the color, either using the active background color or a default color.
-    val backgroundColor = barTintColor ?: getDefaultColorFor(android.R.attr.colorPrimary) ?: return
+    val backgroundColor = color ?: getDefaultColorFor(android.R.attr.colorPrimary) ?: return
 
     // Apply the same color to both active and inactive states
     val colorDrawable = ColorDrawable(backgroundColor)
--- a/android/src/main/java/com/rcttabview/RCTTabView.kt
+++ b/android/src/main/java/com/rcttabview/RCTTabView.kt
@@ -2,10 +2,13 @@ package com.rcttabview
 
 import android.content.Context
 import android.graphics.drawable.BitmapDrawable
+import android.graphics.drawable.ColorDrawable
 import android.graphics.drawable.Drawable
+import android.util.TypedValue
 import android.net.Uri
 import android.view.Choreographer
 import android.view.MenuItem
+import androidx.appcompat.content.res.AppCompatResources
 import com.facebook.common.references.CloseableReference
 import com.facebook.datasource.DataSources
 import com.facebook.drawee.backends.pipeline.Fresco
@@ -25,6 +28,7 @@ class ReactBottomNavigationView(context: Context) : BottomNavigationView(context
   var items: MutableList<TabInfo>? = null
   var onTabSelectedListener: ((WritableMap) -> Unit)? = null
   private var isAnimating = false
+  private var barTintColor : Int? = null
 
   private val layoutCallback = Choreographer.FrameCallback {
     isLayoutEnqueued = false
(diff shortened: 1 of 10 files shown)


ios/TabViewImpl.swift
```diff
@@ -76,6 +77,9 @@ struct TabViewImpl: View {
         UITabBar.appearance().scrollEdgeAppearance = configureAppearance(for: newValue ?? "")
       }
     }
+    .onAppear {
```
okwasniewski, line 80: You should use `.onChange(of: props.barTintColor)` to allow dynamic changing.

ios/TabViewImpl.swift
```diff
@@ -94,6 +98,20 @@ private func configureAppearance(for appearanceType: String) -> UITabBarAppearan
...
+private func updateTabBarAppearance(with barTintColor: UIColor?) {
+    
+  if #available(iOS 15.0, *) {
+    let appearance = UITabBarAppearance()
+      
+    appearance.backgroundColor = barTintColor
+    UITabBar.appearance().standardAppearance = appearance
+    UITabBar.appearance().scrollEdgeAppearance = appearance
+  } else {
+    UITabBar.appearance().barTintColor = barTintColor
```
okwasniewski, line 111: Doesn't it work for newer OS versions? I didn't see deprecation in the docs. 

This function should ideally only call `UITabBar.appearance().barTintColor`. Overriding the `appearance` one more time can cause issues with scrollEdgeApperance.

android/src/main/java/com/rcttabview/RCTTabView.kt
```diff
@@ -25,6 +28,7 @@ class ReactBottomNavigationView(context: Context) : BottomNavigationView(context
   var items: MutableList<TabInfo>? = null
   var onTabSelectedListener: ((WritableMap) -> Unit)? = null
   private var isAnimating = false
+  private var barTintColor : Int? = null
```
okwasniewski, line 31: Why do we need to store this? We can pass the value from setBarTintColor directly to `updateBackgroundColors` method.

okwasniewski (commented):
Overall it looks good! Thanks for working on this, can you double check if it doesn't break any existing appearance props?

ios/TabViewImpl.swift
shubhamguptadream11: Thank you for the suggestion!

I initially tried using .onChange(of: props.barTintColor), but it doesn't seem to update the barTintColor as expected in my case. The color only reflects properly when using .onAppear.

To make sure the `barTintColor` reflects both on view load and during dynamic changes, I combined both .onAppear and .onChange(of:). This approach ensures that the UITabBar appearance is set correctly when the view first appears and updates dynamically when barTintColor changes.

Here is what I am going to change:

```
.onAppear {
    updateTabBarAppearance(with props.barTintColor)
}
.onChange(of: barTintColor) { newColor in
    updateTabBarAppearance(with newColor)
}
```

Is this okay? @okwasniewski

android/src/main/java/com/rcttabview/RCTTabView.kt
shubhamguptadream11: Yeah this makes sense Changing this

ios/TabViewImpl.swift
okwasniewski: If thats the case you can do a custom extension on the view, lets keep simillar logic contained:

```swift
extension View {
  @ViewBuilder
  func tabBarTintColor(enabled: Bool) -> some View {
    self
		.onAppear {
            // Do something
         }
       .onChange(of: barTintColor) { newColor in
    // Do something
      }
  }
  ```

File: ios/TabViewImpl.swift (hunk 2 above)
Comment (shubhamguptadream11) [reply ios/TabViewImpl.swift:111]: I tried with this approach as well: `UITabBar.appearance().barTintColor` on iOS 17.2 simulator.

One wired thing happening here is when I reach end by scrolling till end barTintColor disappeared.

Attaching video for reference:
Simulator Screen Recording - iPhone 15 Pro - 2024-10-18 at 19.51.47

Comment (okwasniewski) [reply ios/TabViewImpl.swift:111]: @shubhamguptadream11 This looks like its related to `scrollEdgeAppearance` (which controls whether the tab bar should disappear when nothing is behind.

Comment (shubhamguptadream11) [reply ios/TabViewImpl.swift:111]: Yess

Comment (shubhamguptadream11) [reply ios/TabViewImpl.swift:111]:
```
  private func updateTabBarAppearance(with barTintColor: UIColor?) {
      if (barTintColor != nil) {
          
          if #available(iOS 15.0, *) {
              let appearance = UITabBarAppearance()
              
              appearance.configureWithOpaqueBackground()
              appearance.backgroundColor = barTintColor
              
              UITabBar.appearance().standardAppearance = appearance
              UITabBar.appearance().scrollEdgeAppearance = appearance
          } else {
              UITabBar.appearance().barTintColor = barTintColor
          }
      }
  }
```


**Why are we putting a check for iOS 15? Is barTintColor deprecated?**

No, barTintColor is not deprecated. The issue I raised above requires setting scrollEdgeAppearance, which is available from iOS 15. To set this, we create a UITabBarAppearance(), making the direct use of barTintColor redundant. Instead, we use backgroundColor.

**Why are we not only using barTintColor to set the tab bar's color?**
This is due to the need to set scrollEdgeAppearance, as mentioned above.

**Why do we check for nil for barTintColor?**
We check for nil to ensure that barTintColor is applied only when the user passes it from the React Native side. This way, scrollEdgeAppearance is overridden only when barTintColor is being applied.

Let me know if you have any other suggestion.

File: android/src/main/java/com/rcttabview/RCTTabView.kt [hunk 4]
```diff
@@ -140,4 +143,28 @@ class ReactBottomNavigationView(context: Context) : BottomNavigationView(context
     super.onDetachedFromWindow()
     isAnimating = false
   }
+
+  fun setBarTintColor(color: Int?) {
+    updateBackgroundColors(color)
+  }
```
okwasniewski, line 149: nit: You can just move the function body here, no need for `updateBackgroundColors` as its not reused anywhere

ios/TabViewImpl.swift
okwasniewski: Oh okay, now I get what you mean. 

In this case we need to have one function which updates everything related to apperance. 

```swift
private func configureAppearance(for appearanceType: String, appearance: UITabBarAppearance) -> UITabBarAppearance {
  switch appearanceType {
  case "opaque":
    appearance.configureWithOpaqueBackground()
  case "transparent":
    appearance.configureWithTransparentBackground()
  default:
    appearance.configureWithDefaultBackground()
  }
  
  return appearance
}

// Helper function to update the tab bar appearance
private func updateTabBarAppearance(props: TabViewProps) {
  if #available(iOS 15.0, *) {
    let appearance = UITabBarAppearance()
    UITabBar.appearance().scrollEdgeAppearance = configureAppearance(
      for: props.scrollEdgeAppearance ?? "",
      appearance: appearance
    )
    appearance.backgroundColor = props.barTintColor
    UITabBar.appearance().standardAppearance = appearance
  } else {
    UITabBar.appearance().barTintColor = props.barTintColor
  }
```

Then we can combine it together with onChange: 

```swift
.onAppear {
      updateTabBarAppearance(props: props)
    }
    .onChange(of: props.barTintColor) { newValue in
      updateTabBarAppearance(props: props)
    }
    .onChange(of: props.scrollEdgeAppearance) { newValue in
      updateTabBarAppearance(props: props)
    }
```

This should also fix current bug with scrollEdgeAppearance because its not updated `onAppear`

okwasniewski: Can you rebase on top of main as I've merged your recent PR, we need to also include it here

shubhamguptadream11: @okwasniewski 
I refactored the code to handle the `onAppear` and `onChange` logic for all TabView props in one place, simplifying the update logic. I’ve tested the changes across all three props—**translucent, scrollEdgeAppearance, and barTintColor**—with all their possible values to ensure everything is working as expected.

okwasniewski (approved):
Thank you for working on this!

PR #62, opened 2024-10-19
feat(iOS): Added translucent property for iOS

shubhamguptadream11:
Part of this: https://github.com/okwasniewski/react-native-bottom-tabs/issues/11

**Summary :** This PR adds the support for translucent in iOS.

Adding video for reference:

`translucent={false}`

https://github.com/user-attachments/assets/6d92a352-bc93-4288-bb3b-cfcaf0d57fe1

`translucent={true}`

https://github.com/user-attachments/assets/66e3b296-59c4-4777-ae20-3b1ac4c340aa

### DIFF
--- a/ios/TabViewImpl.swift
+++ b/ios/TabViewImpl.swift
@@ -15,6 +15,7 @@ class TabViewProps: ObservableObject {
   @Published var ignoresTopSafeArea: Bool?
   @Published var disablePageAnimations: Bool = false
   @Published var scrollEdgeAppearance: String?
+  @Published var translucent: Bool = true
 }
 
 /**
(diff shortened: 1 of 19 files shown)


ios/TabViewImpl.swift
```diff
@@ -155,6 +157,17 @@ extension View {
...
+    @ViewBuilder
+    func tabBarTranslucent(_ translucent: Bool) -> some View {
+        self
+            .onAppear {
+                UITabBar.appearance().isTranslucent = translucent
+            }
+            .onChange(of: translucent) { newValue in
+                UITabBar.appearance().isTranslucent = newValue
+            }
     }
```
okwasniewski, line 172: Looks like you are using 4 space indentation while the project is using 2 spaces, can you fix that? We probably need SwiftLint in the repo to catch this

okwasniewski (commented):
Overall looks good! Thanks for implementing this. Did you try to run it on Android? I think that it needs an empty setter with prop name translucent because without it it might fail

ios/TabViewImpl.swift
shubhamguptadream11: Added an empty setter in android and fixed indentation issues as well.

Thank you for suggestion @okwasniewski. Will keep these in mind in future PRs.

okwasniewski (approved):
Thank you!

Issue #46, opened 2024-10-15
macOS support


okwasniewski:
The goal of this issue is to add macOS support

PR #63, opened 2024-10-19
feat: macos support

okwasniewski:
This PR solves #46 

## Unresolved issue

The view is not sized properly when using sidebar, it's width is always width of the window.

CleanShot 2024-10-19 at 20 32 34@2x

### DIFF
--- a/apps/example/macos/Podfile
+++ b/apps/example/macos/Podfile
@@ -1,0 +1,13 @@
+ws_dir = Pathname.new(__dir__)
+ws_dir = ws_dir.parent until
+  File.exist?("#{ws_dir}/node_modules/react-native-test-app/macos/test_app.rb") ||
+  ws_dir.expand_path.to_s == '/'
+require "#{ws_dir}/node_modules/react-native-test-app/macos/test_app.rb"
+
+workspace 'ReactNativeBottomTabs.xcworkspace'
+
+use_test_app! do |test_app|
+  # Workaround for not using use_frameworks! in the Podfile
+  pod 'SDWebImage', :modular_headers => true
+  pod 'SDWebImageSVGCoder', :modular_headers => true
+end
--- a/.changeset/popular-oranges-provide.md
+++ b/.changeset/popular-oranges-provide.md
@@ -1,0 +1,6 @@
+---
+"react-native-bottom-tabs": patch
+"react-native-bottom-tabs-example": patch
+---
+
+feat: macos support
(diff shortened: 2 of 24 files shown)


okwasniewski:
cc: @Saadnajmi maybe you know what I've missed to make it work with the sidebar? 

I've been also getting these kind of errors, not sure if this is related to RN macOS or my implementation: 

CleanShot 2024-10-19 at 20 46 36@2x

ios/TabViewProvider.swift
```diff
@@ -154,12 +175,23 @@ struct TabData: Codable {
...
+#elseif os(macOS)
+    let newImage = NSImage(size: size)
+    newImage.lockFocus()
+    self.draw(in: CGRect(origin: .zero, size: size),
+              from: CGRect(origin: .zero, size: self.size),
+              operation: .copy,
+              fraction: 1.0)
+    newImage.unlockFocus()
+    return newImage
+#endif
```
Saadnajmi, line 195: See a similar change here: https://github.com/microsoft/react-native-macos/pull/2209
`lockFocus` and that pattern is deprecated in favor of [imageWithSize:flipped:drawingHandler:](https://developer.apple.com/documentation/appkit/nsimage/1519860-imagewithsize). Similarly, I think `UIGraphicsBeginImageContextWithOptions` is mostly replaced with `UIGraphicsImageRenderer`

okwasniewski: Thanks! I will fix it

apps/example/macos/Podfile
```diff
@@ -0,0 +1,13 @@
...
+  File.exist?("#{ws_dir}/node_modules/react-native-test-app/macos/test_app.rb") ||
+  ws_dir.expand_path.to_s == '/'
+require "#{ws_dir}/node_modules/react-native-test-app/macos/test_app.rb"
+
+workspace 'ReactNativeBottomTabs.xcworkspace'
+
+use_test_app! do |test_app|
+  # Workaround for not using use_frameworks! in the Podfile
+  pod 'SDWebImage', :modular_headers => true
+  pod 'SDWebImageSVGCoder', :modular_headers => true
```
Saadnajmi, line 12: Are these because expo-image uses them?

okwasniewski: Bottom tabs use them, it's a workaround for not needing to use static_frameworks - in the installation instructions I recommend to use static frameworks (or people could also enable modular headers globally but I had some issues with it). 

It's about cocoapods and mixing Swift with Objective-C

PR #67, opened 2024-10-21
WIP: Adding barStyle support to control dark mode on android & iOS

shubhamguptadream11:


shubhamguptadream11:
Hi @okwasniewski 

We tried using the standard approach for setting the `barStyle` like this:
```
UITabBar.appearance().barStyle = UIBarStyle.black
```

However, this didn't seem to work as expected.

As an alternative, we’re considering manually implementing it like so:
```
if props.barStyle != nil {
    if props.barStyle == "black" {
        appearance.backgroundColor = UIColor.black
        appearance.selectionIndicatorTintColor = UIColor.white
    } else {
        appearance.backgroundColor = UIColor.white
        appearance.selectionIndicatorTintColor = UIColor.gray
    }
}
```

Would this be a correct approach for manually handling the barStyle behavior, or is there a better way to handle this?

Thanks for your insights!

okwasniewski:
Hey, @shubhamguptadream11 

I think this should work: 

`.preferredColorScheme(.dark)` or `.preferredColorScheme(.light)`

shubhamguptadream11:
@okwasniewski 
It didn't worked, I tried applying this to TabView and Item level both.

I found this: https://www.hackingwithswift.com/forums/swiftui/preferredcolorscheme-not-affecting-datepicker-and-confirmationdialog/11796

As per this it's suggested to use this:
```
private func setupColorScheme(_ isDarkMode: Bool) {
  UIApplication.shared.keyWindow?.overrideUserInterfaceStyle = isDarkMode ? .dark : .light
}
```

With this things are working.

okwasniewski:
@shubhamguptadream11 This will force every native component, to change apperance

I'd like to know if `Appearance.setColorScheme` won't do the same. Maybe we don't need this prop if this can't be forced on the component level?

shubhamguptadream11:
By using `Appearance.setColorScheme('dark')` it enforces dark mode on bottom tab.

okwasniewski:
@shubhamguptadream11 I also tried both approaches with UITabBarAppreance and SwiftUI and nothing seems to work. Let's skip this prop for now as it can be forced with Apperance.setColorScheme()

Thanks for working on this anyways!

shubhamguptadream11:
@okwasniewski Closing this PR.

PR #64, opened 2024-10-20
feat(iOS): new arch

okwasniewski:
This PR implements new architecture for iOS.

PR #68, opened 2024-10-21
feat: new arch building on android

sarthak-d11:


okwasniewski:
Thanks for working on this!

Just FYI, Android will need a custom C++ Shadow node in order to layout the bottom navigation bar properly. 

Here is a reference of this in another library: https://github.com/callstack/react-native-slider/pull/589

We can also split the work (base implementation and Cxx shadow node separately)

sarthak-d11:
Hi @okwasniewski 
I have added the custom C++ shadow node but I am still facing some issues.
The bottom tab bar is not displayed on the screen but the screens are visible.
Here is a screenshot

Screenshot_1729365312

I have tried layout inspector on this and these are the results

Screenshot 2024-10-24 at 12 55 27 AM

Screenshot 2024-10-24 at 1 07 06 AM

If you look at these screenshots we can see that there is some BottomNavigationView is shown in Layout Inspector. But it is not displayed on the screen. Please let me know your thoughts about this

okwasniewski:
Hey, I merged #64, I am looking into this!

okwasniewski:
Hey, 

Huge thanks for working on this! I've started off your work and fixed all the issues here: #79

sarthak-d11:
Hi @okwasniewski, I appreciate your review of this PR; could we discuss the mistake I made regarding the Shadow Node (Was it Event Dispatcher or anything else), as I'm unsure about it, and could you also recommend any resources for further reading(Specially for Custom Shadow Node and CPP code of RN)?

okwasniewski:
> Hi @okwasniewski, I appreciate your review of this PR; could we discuss the mistake I made regarding the Shadow Node (Was it Event Dispatcher or anything else), as I'm unsure about it, and could you also recommend any resources for further reading(Specially for Custom Shadow Node and CPP code of RN)?

Sure! Sorry for not providing you any feedback. 

I've spent a lot of time figuring out whats was wrong you were mostly on point but the main issue was the method signature of the measure call on the ViewManager level as it wasn't called at all. I also made it work for iOS because when you opt-out of codegen passing `interfaceOnly` it's not generated for iOS either.

Regarding the dispatcher I removed it and attached it separately for both architectures as I think this makes things simpler, otherwise it was great work! 

Regarding resources, I don't think there are any... I was mostly going through RN source code and checking how it's done for built-in components 

There aren't many resources

okwasniewski:
I've merged #79 once again thanks for your help!

sarthak-d11:
Thanks for your feed back this is really helpful.

PR #80, opened 2024-10-27
feat: add Apple TV support

### LINKED ISSUE #47: tvOS support
The goal of this issue is to add tvOS support

douglowder:
Fixes #47 .

See https://github.com/douglowder/ExpoRouterTV/tree/rn-bottom-tabs for a demo that works on Apple TV with this change.


ios/TabViewImpl.swift
```diff
@@ -188,9 +188,11 @@ extension View {
   
   @ViewBuilder
   func tabBadge(_ data: String?) -> some View {
-    if #available(iOS 15.0, macOS 15.0, visionOS 2.0, *) {
+    if #available(iOS 15.0, macOS 15.0, visionOS 2.0, tvOS 15.0, *) {
       if let data = data, !data.isEmpty {
+        #if !os(tvOS)
         self.badge(data)
+        #endif
```
okwasniewski, line 195: We need to return `self` either way, this can cause another modifier not to be applied. 

```suggestion
        #if !os(tvOS)
        self.badge(data)
        #else
        self
        #endif
```

douglowder:
```suggestion
        #if !os(tvOS)
        self.badge(data)
        #else
        self
        #endif
```

okwasniewski (approved):
Thank you!

PR #83, opened 2024-10-28
Added activeIndicator color in android

shubhamguptadream11:
Adding `activeIndicator` prop in android.

This props is working with `Material Design 3`.

Screenshot_20241028_163815


okwasniewski:
Thanks for working on this! Can you also add implementation for new architecture? Just merged #79 🙏

Will do a release after this PR lands

shubhamguptadream11:
@okwasniewski  Thanks for sharing the screenshot [here](https://github.com/okwasniewski/react-native-bottom-tabs/issues/11#issuecomment-2439999548) showing the activeIndicator color in use. I'm attempting to implement a similar solution on the Android side by setting `itemActiveIndicatorColor = color`, but I'm not seeing any effect in my example component.

Could you confirm if setting `itemActiveIndicatorColor` alone was sufficient in your setup, or were additional steps needed?

okwasniewski:
@shubhamguptadream11 Can you show me a screenshot? Maybe you are using Material2 component styles? 

You can find info about styling here: https://okwasniewski.github.io/react-native-bottom-tabs/docs/guides/android-native-styling.html

okwasniewski:
Checked out your branch and it's working! 

CleanShot 2024-10-28 at 11 33 37@2x

shubhamguptadream11:
Screenshot_20241028_160936
Here is the snap. I am checking in example of native bottom tabs repo.

okwasniewski:
@shubhamguptadream11 I think this doesn't have any effect on Material2 component design only on Material3

okwasniewski (commented):
Thanks for working on this! 

Everything looks good, can you add this to prop to docs?

shubhamguptadream11:
@okwasniewski updated the doc. Should we also mention it's compatibility with material 3?
Like this

```
activeIndicatorColor <Badge text="android" type="info" />
Sets the color of the tab indicator. This option is only compatible with Material3 themes
```

okwasniewski:
> @okwasniewski updated the doc. Should we also mention it's compatibility with material 3? Like this
> 
> ```
> activeIndicatorColor <Badge text="android" type="info" />
> Sets the color of the tab indicator. This option is only compatible with Material3 themes
> ```

Yeah, let's add it

okwasniewski (approved):
Thanks!

Issue #48, opened 2024-10-15
Expo config plugin 


okwasniewski:
Appearance of the TabBar is controlled by `styles.xml` file on Android.

Goal of this issue is to introduce Expo Config plugin that allows to modify the styles + Change AppTheme to Material3 / MaterialComponents.

Issue #40, opened 2024-10-13
Tab bar looks strange on Android (dark mode?)


fredrikburmester:
<img width="520" alt="Screenshot 2024-10-13 at 13 43 10" src="https://github.com/user-attachments/assets/ea868d64-aa46-4036-bd0a-a891d9d9d422">

Code:
```
<NativeTabs.Screen
        name="(libraries)"
        options={{
          title: "Library",
          tabBarIcon: ({ color, focused, size }) =>
            require("@/assets/icons/server.rack.png"),
        }}
      />
```

okwasniewski:
Hey, thanks for opening the issue.

This is indeed an issue with dark mode. The styles of the TabBar on Android are dictated by `styles.xml` file (Native project file).

We will probably need Expo Config plugin to allow for easy changes. Have you modified this file in any way? Also  is the TabBar white if you switch to white mode?

fredrikburmester:
I have not modified the styles.xml at all. Would be great with an Expo config plugin. 

I will test changing to light mode later today and report back.

fredrikburmester:
I found that changing color props did nothing. Have not had time to test light mode yet.

okwasniewski:
> @okwasniewski there is no way to pass colors as props ?

Not at the moment, this issue #11 is tracking the progress

okwasniewski:
Issue tracking expo config plugin: #48

PR #94, opened 2024-10-29
feat: introduce expo config plugin

okwasniewski:
This PR should close: 

- #48 
- #40 

Api of the config plugin: 

```
"plugins": [
      "react-native-bottom-tabs"
     // or 
     ["react-native-bottom-tabs", { theme: "material2" }]
 ],
```

### DIFF
--- a/app.plugin.js
+++ b/app.plugin.js
@@ -1,0 +1,1 @@
+module.exports = require('./lib/commonjs/expo');
--- a/docs/docs/docs/getting-started/quick-start.mdx
+++ b/docs/docs/docs/getting-started/quick-start.mdx
@@ -6,7 +6,8 @@ import { PackageManagerTabs } from '@theme';
 
 <PackageManagerTabs command="install react-native-bottom-tabs" />
 
-If you use React Native version 0.75 or lower:
+<details>
+  <summary>If you use React Native <b>version 0.75 or lower</b></summary>
 
 - For `@react-native-community/cli` users, open Podfile in ios folder and change minimum iOS version to `14.0` before `pod install`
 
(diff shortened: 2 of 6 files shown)


fredrikburmester:
>  ["react-native-bottom-tabs", { theme: "material2" }]

What themes are there to pick from?

fredrikburmester:
Ah i found this: https://okwasniewski.github.io/react-native-bottom-tabs/docs/guides/android-native-styling.html

fredrikburmester:
Another question, this does not seem to enable the native "pill" look on android. Is that not supported currently in the expo config? Thanks.

okwasniewski:
> Another question, this does not seem to enable the native "pill" look on android. Is that not supported currently in the expo config? Thanks.

The pill is material3 theme which is the default. Just use ["react-native-bottom-tabs"] in the app.json file

fredrikburmester:
Screenshot 2024-11-04 at 22 07 32
I did, but i can't see the pill design.

fredrikburmester:
https://github.com/fredrikburmester/streamyfin/blob/65a6ab99721594f0e9ec4a39070ddebd34057e19/app.json#L104

okwasniewski:
Ah you also added edge-to-edge!

This library is missing material3 style, here is a feature request tracking it: https://github.com/zoontek/react-native-edge-to-edge/issues/38

When it lands you should be able to remove bottom tabs config plugin as only one of them is needed to set the styling

fredrikburmester:
Oh I did not know those were related, sorry. I'm confused as to why the edge-to-edge library needs the _styles_ for the bottom-tabs to look like material3?

okwasniewski:
> Oh I did not know those were related, sorry. I'm confused as to why the edge-to-edge library needs the _styles_ for the bottom-tabs to look like material3? 

Because edge to edge also changes the Android style in styles.xml

Now they change it to AppCompat but we need an option to change it to Material3. Android styles.xml is a bit wired but they base it on inheritance

edmiquilena:
First of all, thanks for the lib!

I'm having the same problem with the "pill" not showing,
The style is shown as material 2
image
image
 "@bottom-tabs/react-navigation": "^0.7.1" with expo SDK 52

Issue #93, opened 2024-10-29
Type errors on React Navigation v7 for createNativeBottomTabNavigator()


shovel-kun:
Seems like they changed the API for `createNavigatorFactory()` and `DefaultNavigatorOptions`

Got type error `Expected 0 type arguments, but got 4.` for the following block of code:
```ts
export default createNavigatorFactory<
  TabNavigationState<ParamListBase>,
  NativeBottomTabNavigationOptions,
  NativeBottomTabNavigationEventMap,
  typeof NativeBottomTabNavigator
>(NativeBottomTabNavigator);
```

As well as type error `Generic type 'DefaultNavigatorOptions' requires 6 type argument(s).` for the block of code:
```ts
export type NativeBottomTabNavigatorProps = DefaultNavigatorOptions<
  ParamListBase,
  TabNavigationState<ParamListBase>,
  NativeBottomTabNavigationOptions,
  NativeBottomTabNavigationEventMap
> &
  TabRouterOptions &
  NativeBottomTabNavigationConfig;
```

Tested with `@react-navigation/native@7.0.0-rc.21`

PR #103, opened 2024-10-31
feat: update createNativeBottomTabsNavigator for React Navigation 7

shovel-kun:
My attempt at fixing the types of createNativeBottomTabsNavigator in React Navigation 7, as brought up #93 

Breaking change, pretty sure this won't work on React Navigation 6.

I referred to the following files for implementation:
- https://github.com/react-navigation/react-navigation/blob/main/packages/bottom-tabs/src/views/BottomTabView.tsx
- https://github.com/okwasniewski/react-native-bottom-tabs/blob/main/src/TabView.tsx
- https://github.com/react-navigation/react-navigation/blob/main/packages/bottom-tabs/src/navigators/createBottomTabNavigator.tsx
- https://github.com/react-navigation/react-navigation/blob/dfb56c62f21728a7e130945fd4e7b74e9c130fff/packages/material-top-tabs/src/views/MaterialTopTabView.tsx#L58 (onIndexTabChange)


okwasniewski:
Hey! Thanks for working on this

FYI there is a guide on how to do this: https://reactnavigation.org/docs/7.x/custom-navigators/

I need to check how can we handle the breaking change

matinzd:
What I suggest is to have another file for v7 so that we keep the v6 support as well. We can keep v6 support by default and import the new react navigation v7 integration with something like:

```
// This will be compatible with v7
import {createNativeBottomTabNavigator} from 'react-native-bottom-tabs/react-navigation/v7';

// This will be compatible with v6
import {createNativeBottomTabNavigator} from 'react-native-bottom-tabs/react-navigation';
```

Or we can do it the other way around and introduce a breaking change version to set v7 as default.

WDYT? @okwasniewski

okwasniewski:
Here is first PR that moves us in the direction of supporting React Navigation v7: 

https://github.com/okwasniewski/react-native-bottom-tabs/pull/152

It introduces a separate package for the integration: `@bottom-tabs/react-navigation` which will be released as 0.6.0, then in 0.7.0 the support for React Navigation v6 will be dropped but people will still be able to use `@bottom-tabs/react-navigation@0.6.0` for React navigation v6, this will allow smoother migration to React Navigation v7 and doesn't bring the maintenance cost of supporting two versions indefinitely

shovel-kun:
Closing as there's a better PR

PR #106, opened 2024-10-31
feat(docs): back behaviour props added for Android

shubhamguptadream11:
Resolves this: https://github.com/okwasniewski/react-native-bottom-tabs/issues/69 
- `backbehaviour` props added in docs.


okwasniewski (approved):
Thank you!

PR #108, opened 2024-10-31
feat: native support for haptic feedback

okwasniewski:
This PR adds native support for haptic feedback!

It uses performHapticFeedback on Android and `UINotificationFeedbackGenerator` on iOS

It's enabled by default and can be disabled by passing `hapticFeedbackEnabled={false}` to the navigator


okwasniewski:
@benkingcode You are right, this should be set to false by default 😅 This can be changed in the next release

PR #113, opened 2024-11-03
Adding freezeOnBlur for android and iOS

shubhamguptadream11:


shubhamguptadream11:
Hey @okwasniewski,

I've implemented the `freezeOnBlur` property, following the same approach as in the JS bottom tabs. I've tested it on **Android**, both with `enableFreeze()` globally and by overriding it through options for individual tabs. Currently, I’m testing this on **iOS** to ensure everything works smoothly there as well.

Meanwhile, feel free to review the PR when you have a chance.

Thanks!

src/TabView.tsx
```diff
@@ -238,39 +260,51 @@ const TabView = <Route extends BaseRoute>({
...
-      {trimmedRoutes.map((route) => {
-        if (getLazy({ route }) !== false && !loaded.includes(route.key)) {
-          // Don't render a screen if we've never navigated to it
-          if (Platform.OS === 'android') {
-            return null;
+      <MaybeScreenContainer
+        enabled={detachInactiveScreens}
+        hasTwoStates
+        style={styles.container}
+      >
```
okwasniewski, line 267: Have you tested it on both new and old architectures? I feel like this might break passing children on new arch as technically now we have only one child.

okwasniewski:
I got instant crash on new arch when going into the example 

CleanShot 2024-11-03 at 18 57 55@2x

src/TabView.tsx
shubhamguptadream11: Till now I tested this on Android for both new and old arch.
I will be testing it on iOS today.

shubhamguptadream11:
@okwasniewski Fixed the iOS crash and tested it on Old and New arch for both platforms. Seems to be working fine now with `react-native-screens` enabled and disabled.

src/react-navigation/views/ScreenFallback.tsx
```diff
@@ -0,0 +1,46 @@
...
+  hasTwoStates: boolean;
+  children?: React.ReactNode;
+}) => {
+  if (Platform.OS === 'android' && Screens?.screensEnabled()) {
+    return (
+      <Screens.ScreenContainer enabled={enabled} {...rest}>
+        {children}
+      </Screens.ScreenContainer>
+    );
+  }
```
okwasniewski, line 36: Looks like we don't have the navigation container on iOS (JS Tabs have it) but it works without it? 

That's weird 😅

If the `ScreenContainer` is required then adding this might not be possible. As we **need** to have multiple children here 

@tboba Could you help here? 🙏

shubhamguptadream11: @okwasniewski Yeah on iOS it works even without Screen Container.

In case of Android we need `ScreenContainer` and I just checked that It worked even with `hasTwoStates={false}` on both platforms with New and Old Arch.

okwasniewski: Hey, @shubhamguptadream11 after consulting with `react-native-screens` team (thanks @kkafar, @tboba)  it looks like `detachInactiveScreens` won't work without the `ScreenContainer`. 

Can you change the PR to only add `freezeOnBlur`?

Nodonisko:
Any updates on this PR? :)

shubhamguptadream11:
Hi @Nodonisko 
Actually I am not available in past few weeks. 
Currently I am working on benchmarking native bottom tabs with JS one. Will pick this PR after that.

okwasniewski:
Merged in #207 thanks @shubhamguptadream11

Issue #124, opened 2024-11-05
createNativeBottomTabNavigator creates truncated labels on iPadOS < 18 and iOS in landscape


johankasperi:
Thanks for an awesome library! This is a huge improvement for React Native.

When using `createNativeBottomTabNavigator` the labels are truncated on iPadOS < 18 and on iOS when in landscape. I can replicate this in the example project. The same bug is not present when using `TabView`. I'm using Xcode 16.1

### iPad 10th generation iPadOS 17.5 (Simulator)
![Simulator Screenshot - iPad (10th generation) - 2024-11-05 at 11 36 15](https://github.com/user-attachments/assets/347cfc39-7d38-4dc1-b1b6-8d3378d1c2d1)

### iPhone 16 Pro iOS 18.1 (Simulator)
![Simulator Screenshot - iPhone 16 Pro - 2024-11-05 at 11 40 43](https://github.com/user-attachments/assets/148a3104-276c-4241-a22e-2f1d4172082a)

I've tried to look into what is causing this but haven't found anything yet.

johankasperi:
It's happening when I'm setting `tabBarInactiveTintColor` (can replicate the bug on `TabView` when setting this property) and it's because of this code https://github.com/okwasniewski/react-native-bottom-tabs/blob/e96c92c2dfe58b04e115e5877dcf24d9fd0cea03/ios/TabViewImpl.swift#L157

okwasniewski:
Thanks for reporting the issue and doing the initial debugging! 

Weird that background color affects this 😅 I will look into this

johankasperi:
Are you sure that we need to do the "hack" suggested in https://stackoverflow.com/a/71934882 ?

I've tested with this code and almost got everything working:
```swift
    UITabBar.appearance().barTintColor = props.barTintColor
    UITabBar.appearance().isTranslucent = props.translucent
    UITabBar.appearance().unselectedItemTintColor = props.inactiveColor
```

But to make that work we cannot override the appearance with:
```swift
  UITabBar.appearance().standardAppearance = appearance
```
so we have to remove that code

okwasniewski:
I am not sure about this, maybe it's not needed. 

If you got it working can you open a PR? 🙏

johankasperi:
> I am not sure about this, maybe it's not needed.
> 
> If you got it working can you open a PR? 🙏

[...]

Issue #128, opened 2024-11-07
shifting label.


a-eid:
I am getting this weird behavior with one of the tabs, where it shifts a bit when clicked, also when clicked when active it shifts sometimes. 

![shifting-label](https://github.com/user-attachments/assets/f22ab5a3-5f0d-45aa-9f55-fa99f8a92114)

Edit: 

this seem to happen because label is too wide!

sometimes it just truncate the label, not always thou. 

<img width="195" alt="Screenshot 2024-11-08 at 12 14 38 AM" src="https://github.com/user-attachments/assets/c5c63fee-b71b-4c2c-ab29-629a380dc85b">

FYI( this happens when I remove the 's' & press on the tab once or twice ).

okwasniewski:
Thanks for reporting the issue, looks like it's related to #124 I'm going to look into this

okwasniewski:
This should be fixed in 0.5.0 when setting `scrollEdgeAppreance="transparent"`. Unfortunately it looks like when setting different apperance the items are jumping which is a bug in Tab View component. This can be reported as a bug for Apple 😢

okwasniewski:
I'm going to close this as there is nothing we can do about SwiftUI bug

PR #133, opened 2024-11-09
fix: appearance handling on iOS

okwasniewski:
This PR should fix: #124 

CleanShot 2024-11-09 at 10 33 02@2x
CleanShot 2024-11-09 at 10 32 44@2x

It's also a partial fix for #128, it appears that everything works properly with `scrollEdgeApperance="transparent"`. 

If we override `tabBar.scrollEdgeAppearance` then it's breaking. It is a bug in SwiftUI's tab view component which we can do nothing about 😢 (other than open a issue for apple)

### DIFF
--- a/ios/Extensions.swift
+++ b/ios/Extensions.swift
@@ -37,3 +37,16 @@ extension UIImage {
     return resizedImage
   }
 }
+
+extension View {
+  @ViewBuilder
+  func introspectTabView(closure: @escaping (UITabBarController) -> Void) -> some View {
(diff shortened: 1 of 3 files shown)


okwasniewski:
This PR also properly encapsulates tab bar styles (instead of app wide styling on UITabBar().appearance) you can now push a new screen with tab bar. 

https://github.com/user-attachments/assets/48b6dcda-1814-4e91-b973-331883d66865

PR #134, opened 2024-11-09
fix(iOS): various fixes for Apple TV

douglowder:
The changes in #116 introduced noticeable rerendering in tvOS when using the arrow key or swipe gesture on the remote to change tabs.

The issue is that on tvOS, tabs are not selected by a press event, so the `onTabItemEvent()` handler is unnecessarily redrawing the tab bar.

This PR fixes that issue, and adds tvOS version checks to the existing iOS version checks.

Also fixed warnings in `PageSelectedEvent.swift`.

### DIFF
--- a/ios/PageSelectedEvent.swift
+++ b/ios/PageSelectedEvent.swift
@@ -27,7 +27,7 @@ import React
   @objc public func arguments() -> [Any] {
     return [
       viewTag,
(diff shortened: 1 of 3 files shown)


okwasniewski (approved):
Thank you! 🙏

PR #140, opened 2024-11-12
fix: tvOS compile error and appearance

douglowder:
- Change to have tab bar correctly appear on first render (otherwise has the default gray color, and won't have correct appearance until tab selection changes)
- Fix compile error (`UIFont.smallSystemFontSize unavailable on tvOS`)

### DIFF
--- a/ios/TabViewImpl.swift
+++ b/ios/TabViewImpl.swift
@@ -125,15 +125,18 @@ struct TabViewImpl: View {
         }
         .tag(tabData?.key)
         .tabBadge(tabData?.badge)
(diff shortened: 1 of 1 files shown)


okwasniewski:
Thanks for this! Looks like the issue with appearance is happening on iOS too, Im going to fix it in separate PR. 

I will also rework the repo structure to be a monorepo so we can fit the Expo example and run tvOS/AndroidTV builds on CI to make sure it doesn't break all the time (sorry for that!)

okwasniewski (approved):
Thanks!

PR #144, opened 2024-11-12
Fix ios build on old xcode

juliesaia-vendora:
Version >=0.5.0 was failing to build on XCode 15.2:

```
func introspectTabView(closure: @escaping (UITabBarController) -> Void) -> some View {
  44 |     self
> 45 |       .introspect(
     |        ^ call to main actor-isolated instance method 'introspect' in a synchronous nonisolated context
```

 I think it's related to a recent change to @MainActor in XCode 16?

Whatever the case, the package is building for me with this change 🙌

### DIFF
--- a/ios/Extensions.swift
+++ b/ios/Extensions.swift
@@ -39,6 +39,7 @@ extension UIImage {
 }
 
 extension View {
+  @MainActor
   @ViewBuilder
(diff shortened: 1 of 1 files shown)


okwasniewski (approved):
Thanks!

Issue #165, opened 2024-11-27
SidebarAdaptable does not push screen content


Robert27:
### Before submitting a new issue

- [X] I tested using the latest version of the library, as the bug might be already fixed.
- [X] I tested using a [supported version](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md) of react native.
- [X] I checked for possible duplicate issues, with possible answers.

### Bug summary

If I set `sidebarAdaptable` to true, then the sidebar superimposes the content of the screen, instead of it being pushed to the side and nothing is covered.
As a enduser I expect the sidebar not to overlay may app content.

### Library version

0.7.0

### Environment info

```shell
System:
  OS: macOS 15.1.1
  CPU: (10) arm64 Apple M1 Pro
  Memory: 260.08 MB / 16.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 23.3.0
    path: /opt/homebrew/bin/node
  Yarn:
    version: 4.0.1
    path: /opt/homebrew/bin/yarn
  npm:
    version: 10.9.0
    path: /opt/homebrew/bin/npm
  Watchman:
    version: 2024.11.18.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.16.2
    path: /Users/roberteggl/.rbenv/shims/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 24.1
      - iOS 18.1
      - macOS 15.1
      - tvOS 18.1
      - visionOS 2.1
      - watchOS 11.1
  Android SDK: Not Found
IDEs:
  Android Studio: 2024.2 AI-242.23339.11.2421.12483815
  Xcode:
    version: 16.1/16B40
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.1
    path: /usr/bin/javac
  Ruby:
    version: 3.2.2
    path: /Users/roberteggl/.rbenv/shims/ruby
npmPackages:
  "@react-native-community/cli":
    installed: 15.1.2
    wanted: latest
  react:
    installed: 18.3.1
    wanted: 18.3.1
  react-native:
    installed: 0.76.3
    wanted: 0.76.3
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: false
iOS:
  hermesEnabled: true
  newArchEnabled: false
```

[...]

Issue #161, opened 2024-11-25
`headerTransparent: false` in embedded stack adds white area at bottom of screen


johankasperi:
### Before submitting a new issue

- [X] I tested using the latest version of the library, as the bug might be already fixed.
- [X] I tested using a [supported version](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md) of react native.
- [X] I checked for possible duplicate issues, with possible answers.

### Bug summary

When using `headerTransparent: false` and `headerLargeTitle: false` in an embedded native stack navigator a white area occurs at bottom of screen on iPad with iOS >= 18. The bug can be replicated in the example app when changing this line to `false` https://github.com/okwasniewski/react-native-bottom-tabs/blob/af0c62af8ab408a6b5e78622dca32b01b774aecc/apps/example/src/Examples/NativeBottomTabsEmbeddedStacks.tsx#L10

Printscreen of the bug:
![Simulator Screenshot - iPad Air 11-inch (M2) - 2024-11-25 at 16 15 13](https://github.com/user-attachments/assets/c783bfde-e363-48c6-8b74-b317601662ae)

Unsure if the issue is with this library or `react-native-screens` but posting it here since the bug does not occur with @react-navigation/bottom-tabs

### Library version

0.6.0

### Environment info

```shell
From the example app:

System:
  OS: macOS 14.7.1
  CPU: (10) arm64 Apple M1 Pro
  Memory: 297.39 MB / 32.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 20.17.0
    path: ~/.nvm/versions/node/v20.17.0/bin/node
  Yarn:
    version: 3.6.1
    path: ~/.nvm/versions/node/v20.17.0/bin/yarn
  npm:
    version: 10.8.2
    path: ~/.nvm/versions/node/v20.17.0/bin/npm
  Watchman:
    version: 2024.08.12.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.15.2
    path: /Users/johan.kasperi/.rbenv/shims/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 24.1
      - iOS 18.1
      - macOS 15.1
      - tvOS 18.1
      - visionOS 2.1
      - watchOS 11.1
  Android SDK:
    API Levels:
      - "28"
      - "30"
      - "34"
      - "35"
    Build Tools:
[...]

PR #170, opened 2024-12-01
fix: measure available space for views on iOS

okwasniewski:
## PR Description

This PR introduces measuring available space for views so that they get laid out properly. This is internal only change that doesnt affect public API.

This should also unblock #46 

Should fix: 

- #165
- #161 

## How to test?

Check above issues after applying those changes.

## Screenshots

N/A

### DIFF
--- a/.changeset/hungry-readers-remember.md
+++ b/.changeset/hungry-readers-remember.md
@@ -1,0 +1,5 @@
+---
+"react-native-bottom-tabs": patch
+---
+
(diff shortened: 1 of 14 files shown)


okwasniewski:
Next release should bring much better support for `sidebarAdaptable`

https://github.com/user-attachments/assets/7a295b28-10fe-4630-ac17-a2e20f36041f

PR #172, opened 2024-12-02
fix: android tabBarIcon not visible when build to apk

Peek-A-Booo:
## PR Description

I found that the image processing on Android uses [Coil](https://github.com/coil-kt/coil). 

https://github.com/okwasniewski/react-native-bottom-tabs/blob/fc096decd22603ab92c2da290f1dd4f1a2b774d4/packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabView.kt#L184-L198

- When in expo dev, assets are served over HTTP (According to [expo Assets](https://docs.expo.dev/develop/user-interface/assets/#how-are-assets-served-locally)), the `imageSource.uri` looks like `http://localhost:8081/assets/xxxx` and Coil handles it well
- When build to apk, the `imageSource.uri` looks like `res:/2131165303`, Coil cannot handle this and will cause error

Need format

## How to test?

Test on my Android device

### Before
image

### After
image

### DIFF
--- a/packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabView.kt
+++ b/packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabView.kt
@@ -6,6 +6,7 @@ import android.content.res.ColorStateList
 import android.graphics.Typeface
 import android.graphics.drawable.ColorDrawable
 import android.graphics.drawable.Drawable
+import android.net.Uri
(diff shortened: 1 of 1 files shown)


okwasniewski:
Created a follow-up PR #175 based on your work this should handle more cases, in my testing the fix didn't work for every case.

okwasniewski:
Thanks for your work on this one! Should be out in next release

PR #174, opened 2024-12-02
Fix a bug that active tab label shows with inactive color

honcon:
In scrollEdgeAppearance : 'transparent' mode the active tab appear its label with inactive tint color


okwasniewski (approved):
Looks good, thanks!

PR #177, opened 2024-12-03
fix(docs): wrong default value for labeled prop

Nodonisko:
## PR Description

Default value is wrong because labels are by default hidden on Android when tab is not focused.


okwasniewski (commented):
Thanks for the PR! 

Can you also update `guides/standalone-usage.md`?

Nodonisko:
@okwasniewski I actually found out that it's not easy, it's default `false` on Android, but default `true` on iOS which makes sense as it is platform default. I tried to reflect this in docs.

okwasniewski (approved):
Yeah, this makes sense now! Thanks

PR #180, opened 2024-12-04
feat: add support for top border color on iOS

Nodonisko:
## PR Description

It is possible to customize border color of tab bar on iOS.

## Screenshots

Simulator Screenshot - iPhone 16 Pro - 2024-12-04 at 18 54 43


okwasniewski (changes requested):
Thanks for working on this! 

But I'm not sure about this prop. I don't want the interface of TabView to become bloated with too many props. Maybe we can define a `tabBarStyle` prop that will have a limited amount of available properties that people can set. In this case it would be: 

```tsx
<TabView tabBarStyle={{ borderColor: 'red'  }} /> 
```

This way we will follow JS bottom tabs spec and don't end up in 20 props that people need to set on the root component. You are also setting `borderColor` to `shadowColor` Shoudn't this be a separate prop? 

Also, did you check if this is possible on Android?

Nodonisko:
I think adding property just for styles is good idea, but in that case should we move also other color styling properties under this new one?

It should be possible for Android but it won't be that easy as for iOS since by default Android doesn't have any border. Not sure why it's called shadow on iOS because in my opinion it looks like standart border.

okwasniewski:
@Nodonisko the rest of color styling match JS Bottom Tabs so I don't think thats necessary. 

We can start off with only having this style available for iOS

PR #181, opened 2024-12-05
fix: replace RN babel preset with codegen plugin directly

jbroma:
## PR Description

Replaced RN babel preset in the main package with codegen plugin in order to prevent injecting `@babel/runtime` into the code when building the release version. 

This was causing issues in setup where `swc` is used instead of `babel` since `swc` has it's own helpers (equivalent of `@babel/runtime`). I believe helpers should be injected when bundling the project and not at the time of building the library, so there is no hard requirement of having `@babel/runtime` as a dependency in the project.

## How to test?

- [x] - example works
- [ ] - no warning from codegen outside the example apps

## Screenshots

n/a

### DIFF
--- a/.changeset/ninety-planets-sleep.md
+++ b/.changeset/ninety-planets-sleep.md
@@ -1,0 +1,5 @@
+---
+'react-native-bottom-tabs': patch
+---
+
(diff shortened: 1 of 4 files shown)


okwasniewski (approved):
Thank you! 🙏

PR #234, opened 2025-01-09
feat: update gh actions from `v3` to `v4`

krozniata:
## PR Description

Updates few GH actions from `v3` to `v4`, because `v3` use node 16 under the hood which is deprecated

Also updates `.nvmrc` to use `LTS`

## How to test?

Green CI


okwasniewski (approved):
Thank you 🙏

PR #241, opened 2025-01-14
chore: remove tsc path from package.json

krozniata:
## PR Description

Removes hardcoded `tsc` path from `package.json`, as without it it is still being resolved correctly and packages build as expected

## How to test?

Green CI, `yarn build` works correctly


okwasniewski (approved):
Thank you, this is a left-over from my monorepo migration 😅

Issue #248, opened 2025-01-20
Allow changing tabBarLabel or title during runtime


okwasniewski:
### Discussed in https://github.com/callstackincubator/react-native-bottom-tabs/discussions/246

<div type='discussions-op-text'>

<sup>Originally posted by **SumitR9910** January 17, 2025</sup>
Currently it is not possible to change the tabBarLabel or title during runtime, changing the title on runtime throws this error: 

<img width="1148" alt="Screenshot 2025-01-18 at 12 38 25 AM" src="https://github.com/user-attachments/assets/77967962-405d-496a-a144-a45ebb4e7531" />

Implementation of this functionality will enable performing tasks like changing the language of the tab bar label based on the set app language.
</div>

PR #250, opened 2025-01-21
fix: fix crash when change tab bar label during runtime

owinter86:
## PR Description
This fixes the issue where android will crash when changing the tab bar label during run time crashes the app.

Fixes https://github.com/callstackincubator/react-native-bottom-tabs/issues/248

## How to test?

I have added a state change in the longPress event within the NativeBottomTabs example, this should change the first tab from "Article" to  "CHANGED".

### DIFF
--- a/apps/example/src/Examples/NativeBottomTabs.tsx
+++ b/apps/example/src/Examples/NativeBottomTabs.tsx
@@ -4,10 +4,12 @@ import { Contacts } from '../Screens/Contacts';
 import { Chat } from '../Screens/Chat';
 import { createNativeBottomTabNavigator } from '@bottom-tabs/react-navigation';
 import { Platform } from 'react-native';
+import { useState } from 'react';
 
 const Tab = createNativeBottomTabNavigator();
 
 function NativeBottomTabs() {
+  const [label, setLabel] = useState('Article');
   return (
     <Tab.Navigator
       initialRouteName="Chat"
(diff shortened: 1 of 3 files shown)


apps/example/src/Examples/NativeBottomTabs.tsx
```diff
@@ -40,11 +42,13 @@ function NativeBottomTabs() {
             console.log(
               `${Platform.OS}: Long press detected on tab with key ${data.target} at the screen level.`
             );
+            setLabel("CHANGED")
```
okwasniewski, line 45:
```suggestion
            setLabel('New Article')
```

packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabView.kt
```diff
@@ -197,16 +198,17 @@ class ReactBottomNavigationView(context: ReactContext) : LinearLayout(context) {
   }
 
   private fun onTabSelected(item: MenuItem) {
-    val selectedItem = items.first { it.title == item.title }
+    val selectedItem = items[item.itemId]
     selectedItem.let {
       onTabSelectedListener?.invoke(selectedItem.key)
       emitHapticFeedback(HapticFeedbackConstants.CONTEXT_CLICK)
     }
+
```
okwasniewski, line 206:
```suggestion
```

packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabView.kt
```diff
@@ -84,6 +84,7 @@ class ReactBottomNavigationView(context: ReactContext) : LinearLayout(context) {
       LayoutParams.WRAP_CONTENT
     ))
 
+
```
okwasniewski, line 87:
```suggestion
```

okwasniewski (approved):
Thanks for fixing this! 🎉

SumitR9910:
Thanks for the quick PR. Can confirm it works fine now.

PR #251, opened 2025-01-21
fix: fix tabLabelStyle not applying on android

owinter86:
## PR Description

tabLabelStyle is not applied on android currently in the latest release, this fixes this issue so that custom font and font size changes will correctly apply.

## How to test?

Change fontSize in the tabLabelStyle within the NativeBottomTabs route, pre this commit it did not do anything.

### DIFF
--- a/packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabView.kt
+++ b/packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabView.kt
@@ -370,7 +370,7 @@ class ReactBottomNavigationView(context: ReactContext) : LinearLayout(context) {
 
   private fun updateTextAppearance() {
     if (fontSize != null || fontFamily != null || fontWeight != null) {
-      val menuView = getChildAt(0) as? ViewGroup ?: return
+      val menuView = bottomNavigation.getChildAt(0) as? ViewGroup ?: return
       val size = fontSize?.toFloat()?.takeIf { it > 0 } ?: 12f
       val typeface = ReactFontManager.getInstance().getTypeface(
         fontFamily ?: "",


owinter86:
I also notice that the current react font manager instance only allows for font weight changes from bold or normal, other weights will default back to normal.

https://github.com/callstackincubator/react-native-bottom-tabs/blob/c19fdefb4730f4c67fe6a04bf25fb90828a0533f/packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabView.kt#L375-L379

however I also see that its possible to utilise the range of fontWeights within the ReactFontManager class 
```java
public Typeface getTypeface(String fontFamilyName, int weight, boolean italic, AssetManager assetManager) {
  return this.getTypeface(fontFamilyName, new TypefaceStyle(weight, italic), assetManager);
}
```

I tested this locally and was able to have various font weights ranging from 100 to 900 (depending on the font support)
```kotlin
val typeface = ReactFontManager.getInstance().getTypeface(
  fontFamily ?: "",
  fontWeight ?: 400,
  false,
  context.assets
)
```

packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabView.kt
```diff
@@ -370,7 +370,7 @@ class ReactBottomNavigationView(context: ReactContext) : LinearLayout(context) {
 
   private fun updateTextAppearance() {
     if (fontSize != null || fontFamily != null || fontWeight != null) {
-      val menuView = getChildAt(0) as? ViewGroup ?: return
```
okwasniewski, line 373: Thanks for creating this PR, looks like this change is already included in #250

okwasniewski:
> I also notice that the current react font manager instance only allows for font weight changes from bold or normal, other weights will default back to normal.
> 
> https://github.com/callstackincubator/react-native-bottom-tabs/blob/c19fdefb4730f4c67fe6a04bf25fb90828a0533f/packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabView.kt#L375-L379
> 
> however I also see that its possible to utilise the range of fontWeights within the ReactFontManager class
> 
> ```java
> public Typeface getTypeface(String fontFamilyName, int weight, boolean italic, AssetManager assetManager) {
>   return this.getTypeface(fontFamilyName, new TypefaceStyle(weight, italic), assetManager);
> }
> ```
> 
> I tested this locally and was able to have various font weights ranging from 100 to 900 (depending on the font support)
> 
> ```kotlin
> val typeface = ReactFontManager.getInstance().getTypeface(
>   fontFamily ?: "",
>   fontWeight ?: 400,
>   false,
>   context.assets
> )
> ```
> 
> not sure if there were any reason behind the restricted font weight selections, and think its probably outside the scope of this quick fix. But maybe worth adding to this, or a different PR.

You are welcome to create a PR for this support, It seemed off to me when I implemented this and forgot to get back to this later 😅

owinter86:
@okwasniewski ok I will update this to include the additional font weight supports.

PR #253, opened 2025-01-22
fix: fix crash on iOS when adding fontWeight on old arch builds

owinter86:
## PR Description

Fixes crash when adding fontWeight as a label style on iOS, when built using old architecture

## How to test?

Create build with new arch disabled, and add fontWeight as a label style.

## Screenshots

Simulator Screenshot - iPhone 16 Pro - 2025-01-22 at 21 00 34


okwasniewski (approved):
Good catch, thank you for fixing this 🙏

PR #291, opened 2025-02-06
fix: properly handle labeled false on iOS

okwasniewski:
## PR Description

This PR properly handles labeled false on iOS


okwasniewski:
!canary

PR #292, opened 2025-02-06
fix: upgrade material to alpha10 so dynamic tabBarIcon works on Material 3

johankasperi:
When using a app theme with a Material 3 parent theme a dynamic `tabBarIcon` based on `focused` stopped working in version 0.8.0. This change fixes this

### DIFF
--- a/.changeset/funny-oranges-applaud.md
+++ b/.changeset/funny-oranges-applaud.md
@@ -1,0 +1,5 @@
+---
+'react-native-bottom-tabs': patch
+---
+
+chore: bump android material components package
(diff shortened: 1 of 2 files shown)


johankasperi:
After my tests it seems like the bug reappears after opening a modal. But this change at least makes it a bit better. Earlier it was completely impossible to change `tabBarIcon` during runtime with Material 3

okwasniewski:
Thanks for this PR! If you are still running into issues can you describe it in an Issue?

okwasniewski (approved):

johankasperi:
> Thanks for this PR! If you are still running into issues can you describe it in an Issue?

Thank you! Created an issue describing the bug here: https://github.com/callstackincubator/react-native-bottom-tabs/issues/298

PR #299, opened 2025-02-10
fix: remove video url from packages's `README.md`

szymonrybczak:
## PR Description

Remove video url from the `package`'s README.md file because npm don't render it properly

CleanShot 2025-02-10 at 14 57 48@2x

There's another README.md in the root of this directory which still will contain nice looking video :D 

## How to test?

n/a

## Screenshots

n/a


okwasniewski (approved):
Thanks!

Issue #71, opened 2024-10-21
Implement `freezeOnBlur`


okwasniewski:
### `freezeOnBlur`

Boolean indicating whether to prevent inactive screens from re-rendering. Defaults to false. Defaults to true when enableFreeze() from react-native-screens package is run at the top of the application.

Requires react-native-screens version >=3.16.0.

Only supported on iOS and Android.

shubhamguptadream11:
Hi @okwasniewski ,

I've been reviewing the `freezeOnBlur` implementation in the JS bottom tabs. In that setup, `freezeOnBlur` is passed as a `screenOptions` from Tab.navigator or directly in the `options` for each Screen. It ultimately propagates this prop to Screens from `react-native-screens` ([code reference](https://github.com/react-navigation/react-navigation/blob/581681aea465b871728b9d7e4b1a6eebbcf05ed0/packages/drawer/src/views/ScreenFallback.tsx#L18)).

This approach leverages `react-freeze` to manage screen freezing.

However, for **NativeBottomTabs**, we're using platform-specific APIs to create the bottom tabs, which means we aren’t leveraging `react-native-screens`. Given this, would you recommend a custom approach to achieve similar functionality for `freezeOnBlur`?

One possibility might be to use information about loaded tabs to manage freezing behaviour ([see reference](https://github.com/okwasniewski/react-native-bottom-tabs/blob/9ab4eead81ad0820128a09deea4cc82614715756/src/TabView.tsx#L242)).

Additionally, `enableFreeze()` currently has no effect in `NativeBottomTabs`, I’m curious if you’d suggest any alternative approaches or improvements here.

Thanks for your guidance!

okwasniewski:
Hey,

Thanks for looking into this!

I was thinking that we can take the JS implementation and wrap each screen with `ScreenFallback`.

https://github.com/react-navigation/react-navigation/blob/581681aea465b871728b9d7e4b1a6eebbcf05ed0/packages/bottom-tabs/src/views/ScreenFallback.tsx

We are currently wrapping each screen with View anyways so this can be changed to the ScreenFallback: 

https://github.com/okwasniewski/react-native-bottom-tabs/blob/9ab4eead81ad0820128a09deea4cc82614715756/src/TabView.tsx#L257

This should be an optional peer dependency as this prop will be optional.

okwasniewski:
Looks like in V7 its called `detachInactiveScreens`

[...]

PR #302, opened 2025-02-13
feat: implement freezeOnBlur

okwasniewski:
## PR Description

This PR implements freezeOnBlur by using `react-freeze` directly. Should close: #71 

## How to test?

1. Enable freezeOnBlur
2. Check if setTimeout is not called

## Screenshots

N/A

### DIFF
--- a/packages/react-navigation/src/views/NativeBottomTabView.tsx
+++ b/packages/react-navigation/src/views/NativeBottomTabView.tsx
@@ -63,6 +63,9 @@ export default function NativeBottomTabView({
         return null;
       }}
       getLazy={({ route }) => descriptors[route.key]?.options.lazy ?? true}
+      getFreezeOnBlur={({ route }) =>
+        descriptors[route.key]?.options.freezeOnBlur
+      }
       onTabLongPress={(index) => {
         const route = state.routes[index];
         if (!route) {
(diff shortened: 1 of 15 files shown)


packages/react-navigation/src/views/NativeBottomTabView.tsx
```diff
@@ -63,6 +63,9 @@ export default function NativeBottomTabView({
         return null;
       }}
       getLazy={({ route }) => descriptors[route.key]?.options.lazy ?? true}
+      getFreezeOnBlur={({ route }) =>
+        descriptors[route.key]?.options.freezeOnBlur
+      }
```
emzet93, line 68: Does it mean that `freezeOnBlur` needs to be set directly on navigator and it won't be automatically propagated with `enableFreeze(true)` from `react-native-screens` is called?

I guess it's not a big deal - asking out of curiosity!

okwasniewski: Yeah, I don't want to introduce peer dependency on react-native-screens to achieve this small thing, I will state in the docs that this has to be enabled separately.

emzet93: Makes sense, thanks for clarification!

PR #312, opened 2025-02-20
fix: add tvOS 16.0 availability check for toolbar API

mtshv:
Problem:
The toolbar API availability check only includes iOS 16.0, causing build errors on tvOS targets. This was introduced by https://github.com/callstackincubator/react-native-bottom-tabs/commit/024e92e2708eb076079d9eedddb79eeae4c2af9e#diff-89822b2e0d1364951af03c6bbf9b4e69b7e1f1ceca81ad7b242d48ba2195e5a3R355-R368

image

Solution:
Added tvOS 16.0 to the availability check to support both platforms.


fredrikburmester:
Sorry if this is the wrong place to post this. How do I get past this build error?
Screenshot 2025-02-21 at 19 19 23

PR #314, opened 2025-02-21
Provide alternative to enabling static framework on Expo using expo-build-properties

elsa17z:
Current docs specify static framework linking must be enabled for installation, however it is possible to just use expo-build-properties to specify the 2 pods that require modular_headers support to be installed as extra pods (which was the reason why static framework linking had to be enabled in the first place). 

The extraPods option appears to implement essentially the code in the following podfile without having to enable static linking in Expo 

https://github.com/callstackincubator/react-native-bottom-tabs/blob/1b0ed94579bb234afc047a0bae7f2463cb125de0/example/ios/Podfile

Docs update

Just do a fresh new build

### DIFF
--- a/docs/docs/docs/getting-started/quick-start.mdx
+++ b/docs/docs/docs/getting-started/quick-start.mdx
@@ -84,6 +84,28 @@ You also need to enable static linking for iOS by adding `"useFrameworks": "stat
 }
 ```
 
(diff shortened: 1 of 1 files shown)


okwasniewski (approved):
Thanks!

Issue #143, opened 2024-11-12
SVGs don't work with TabBarIcon using it with expo router


boblitex:
while PNGs work as TabBarIcon, SVGs are blank as opposed to what's shown in the documentation ( for react navigation) but you get the idea

okwasniewski:
Hey @boblitex can you provide a minimal reproduction? 

Here I have an example of using SVGs with bottom tabs: https://github.com/okwasniewski/ExpoNativeTabs 

Can you try if it works for you?

okwasniewski:
Can you check with 0.5.2?

Peek-A-Booo:
> Can you check with 0.5.2?

0.5.2 doesn't work for me. 
(only `sfSymbol` and `uri` working, `require` not)

```
<Tab.Screen
  name="Albums"
  component={Albums}
  options={{
    tabBarIcon: () => require('person.png'),
    // SVG is also supported
    tabBarIcon: () => require('person.svg'),
    // or
    tabBarIcon: () => ({ sfSymbol: 'person' }),
    // You can also pass a URL
    tabBarIcon: () => ({ uri: 'https://example.com/icon.png' }),
  }}
/>
```

okwasniewski:
@Peek-A-Booo Can you share code snippet that doesn't work for you?

Just to be sure did you provide a full path? `tabBarIcon: () => require('../../assets/images/home-icon-silhouette.svg') `

Peek-A-Booo:
@okwasniewski code&result with expo router

### Code
<img width="2051" alt="image" src="https://github.com/user-attachments/assets/fb1531e1-f108-4670-80ec-3c854f6d276f">

### Result
![image](https://github.com/user-attachments/assets/93fe03db-7a6c-47c6-9e87-ff1ed946c5ee)

okwasniewski:
@Peek-A-Booo Thanks, and is it working for you on iOS?

Peek-A-Booo:
@okwasniewski Thanks. For me, the only ones working in iOS are `sfSymbol` and `uri`.

Looks like

```
 tabBarIcon: () => ({ sfSymbol: 'person' }),

 tabBarIcon: () => ({ uri: 'https://example.com/icon.png' }),
```

Peek-A-Booo:
@okwasniewski Found my mistake. I introduced `react-native-svg-transformer` to use SVG, and configured it like this:
[...]

PR #362, opened 2025-06-13
docs: add note about transformers like react-native-svg-transformer

Armster15:
## PR Description

`react-native-svg-transformer` is a popular transformer that plenty of people in the RN ecosystem use to import SVGs as React components. This transformer however, breaks SVG usage in this library.

This is not the fault of this library, rather more so on the transformer as it's modifying the default behavior of React Native's (more specifically Metro's) asset handling.

However, as this approach is quite popular, it makes sense adding a note in the docs about it. There's two parts to the warning: first, informing the end user that such transformers will break this functionality. I also added an example of how to resolve this with `react-native-svg-transformer` specifically as it's quite a popular lib. `react-native-svg-transformer` unfortunately overrides the `svg` extension specifically and does not give you an option to change the extension it's overriding (so no `svg?react` for example 😞), so my proposed solution uses another extension that resolves the SVG to Metro's standard resolution of SVGs and uses that in the code for the `tabBarIcon`.

Existing GitHub issue showing people struggling with this: https://github.com/callstackincubator/react-native-bottom-tabs/issues/143#issuecomment-2479561808

Cheers, and thanks for this amazing library!

## How to test?

N/A

## Screenshots

N/A

### DIFF
--- a/docs/docs/docs/guides/usage-with-react-navigation.mdx
+++ b/docs/docs/docs/guides/usage-with-react-navigation.mdx
@@ -244,6 +244,39 @@ Function that given `{ focused: boolean }` returns `ImageSource` or `AppleIcon` 
 SF Symbols are only supported on Apple platforms.
 :::
 
+:::warning
+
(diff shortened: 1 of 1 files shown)


okwasniewski (approved):
Thanks for adding this!

PR #364, opened 2025-06-14
feat: add Tab api support on newer OS

47PADO47:
## PR Description

This pr adds support for the new swift [Tab](https://developer.apple.com/documentation/swiftui/tab) api on ios/ipadOS/tvOS 18+, macOS 15+ and visionOS 2+. This let us implement many new features (such as [search role](https://developer.apple.com/documentation/swiftui/tabrole/search) for the tab item, [tabbar customization](https://developer.apple.com/documentation/swiftui/tabcontent/customizationbehavior(_:for:)), [swipe actions](https://developer.apple.com/documentation/swiftui/tabcontent/swipeactions(edge:allowsfullswipe:content:)) and [context menus](https://developer.apple.com/documentation/swiftui/tabcontent/contextmenu(menuitems:))), while also opening up an opportunity to implement https://github.com/callstackincubator/react-native-bottom-tabs/discussions/296 in the future.

---

In order to implement it i had to create a separate file for each implementation because only by doing so and using `@available(iOS 18, macOS 15, visionOS 2, tvOS 18, *)` on `NewTabView` i could then add  `@AppStorage("sidebarCustomizations")`. I tried to maintain every feature on each implementation and moved the `TabViewImpl#renderTabItem` to `LegacyTabView#renderTabItem`. i was hoping to add a similar function to the NewTabView struct but in that case the Tab api gave an error about not conforming to View (strange since the same code works directly in the ForEach).

As of now i have disabled customization for each Tab (using `.customizationBehavior(.disabled, for: .sidebar, .tabBar)`), as we can then discuss in a followup pr/issue the js implementation, set Tab role to nil (same reason), disabled badges on the new tab (commented line) because even if `tabData.badge` is an empty string, an empty red dot is created, and i could not figure out how to circumvent this. Finally, i found a strange behaviour with the `Native Bottom Tabs with Custom Tab Bar` example, where for the new tab implementation, the native tab is still visible, but even there i could not figure out why.

Let me know your thoughts and if you have any suggestion on how to resolve these issues

## How to test?

Open the app and test the examples

## Screenshots

N/A, the look is the same, only a new edit button has appeared in the ipad/macos sidebar

### DIFF
--- a/packages/react-native-bottom-tabs/ios/TabView/LegacyTabView.swift
+++ b/packages/react-native-bottom-tabs/ios/TabView/LegacyTabView.swift
@@ -1,0 +1,64 @@
+import SwiftUI
+
+struct LegacyTabView: View {
+  @ObservedObject var props: TabViewProps
+
+#if os(macOS)
+  var tabBar: NSTabView?
+#else
+  var tabBar: UITabBar?
+#endif
+  
+  var onLayout: (_ size: CGSize) -> Void
+  var onSelect: (_ key: String) -> Void
+  var updateTabBarAppearance: (_ props: TabViewProps, _ tabBar: UITabBar?) -> Void
+  
+  @ViewBuilder
+  var body: some View {
+    TabView(selection: $props.selectedPage) {
+      ForEach(props.children.indices, id: \.self) { index in
+        renderTabItem(at: index)
+      }
+      .measureView { size in
+        onLayout(size)
+      }
+    }
+    .hideTabBar(props.tabBarHidden)
+  }
+  
+  @ViewBuilder
+  private func renderTabItem(at index: Int) -> some View {
+    if let tabData = props.items[safe: index] {
+      let isFocused = props.selectedPage == tabData.key
+      
+      if !tabData.hidden || isFocused {
+        let icon = props.icons[index]
+        let child = props.children[safe: index] ?? PlatformView()
+        
+        RepresentableView(view: child)
+          .ignoresSafeArea(.container, edges: .all)
+          .tabItem {
+            TabItem(
+              title: tabData.title,
+              icon: icon,
+              sfSymbol: tabData.sfSymbol,
+              labeled: props.labeled
+            )
+            .accessibilityIdentifier(tabData.testID ?? "")
+            .tag(tabData.key)
+            .tabBadge(tabData.badge)
+            .onAppear {
+#if !os(macOS)
+              updateTabBarAppearance(props, tabBar)
+#endif
+#if os(iOS)
+              if index >= 4, !isFocused {
+                onSelect(tabData.key)
+              }
+#endif
+            }
+          }
+      }
+    }
+  }
+}
--- a/packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
+++ b/packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
@@ -1,0 +1,69 @@
+import SwiftUI
+
+@available(iOS 18, macOS 15, visionOS 2, tvOS 18, *)
+struct NewTabView: View {
+  @ObservedObject var props: TabViewProps
+  @AppStorage("sidebarCustomizations") var tabViewCustomization: TabViewCustomization
+
+  #if os(macOS)
+    var tabBar: NSTabView?
+  #else
+    var tabBar: UITabBar?
+  #endif
+
+  var onLayout: (_ size: CGSize) -> Void
+  var onSelect: (_ key: String) -> Void
+  var updateTabBarAppearance: (_ props: TabViewProps, _ tabBar: UITabBar?) -> Void
+
+  @ViewBuilder
+  var body: some View {
+    TabView(selection: $props.selectedPage) {
+      ForEach(props.children.indices, id: \.self) { index in
+        if let tabData = props.items[safe: index] {
+          let isFocused = props.selectedPage == tabData.key
+
+          if !tabData.hidden || isFocused {
+            let icon = props.icons[index]
+            let role: TabRole? = nil
+
+            let platformChild = props.children[safe: index] ?? PlatformView()
+            let child = RepresentableView(view: platformChild)
+
+            Tab(value: tabData.key, role: role) {
+              child
+                .ignoresSafeArea(.container, edges: .all)
+                .onAppear {
+                  #if !os(macOS)
+                    updateTabBarAppearance(props, tabBar)
+                  #endif
+                  #if os(iOS)
+                    if index >= 4 {
+                      if props.selectedPage != tabData.key {
+                        onSelect(tabData.key)
+                      }
+                    }
+                  #endif
+                }
+            } label: {
+              TabItem(
+                title: tabData.title,
+                icon: icon,
+                sfSymbol: tabData.sfSymbol,
+                labeled: props.labeled
+              )
+            }
+            //.badge(tabData.badge)
+            .customizationID(tabData.key)
+            .customizationBehavior(.disabled, for: .sidebar, .tabBar)
+            .accessibilityIdentifier(tabData.testID ?? "")
+          }
+        }
+      }
+    }
+    .measureView { size in
+      onLayout(size)
+    }
+    .tabViewCustomization($tabViewCustomization)
+    .hideTabBar(props.tabBarHidden)
+  }
+}
--- a/packages/react-native-bottom-tabs/ios/TabViewImpl.swift
+++ b/packages/react-native-bottom-tabs/ios/TabViewImpl.swift
@@ -13,21 +13,27 @@ struct TabViewImpl: View {
 #else
   @Weak var tabBar: UITabBar?
 #endif
-
+  
+  @ViewBuilder
+  var tabContent: some View {
+    if #available(iOS 18, macOS 15, visionOS 2, tvOS 18, *) {
+      NewTabView(
+        props: props, tabBar: tabBar, onLayout: onLayout, onSelect: onSelect,
+        updateTabBarAppearance: updateTabBarAppearance)
+    } else {
+      LegacyTabView(
+        props: props, tabBar: tabBar, onLayout: onLayout, onSelect: onSelect,
+        updateTabBarAppearance: updateTabBarAppearance)
+    }
+  }
+  
   var onSelect: (_ key: String) -> Void
   var onLongPress: (_ key: String) -> Void
   var onLayout: (_ size: CGSize) -> Void
   var onTabBarMeasured: (_ height: Int) -> Void
-
+  
   var body: some View {
-    TabView(selection: $props.selectedPage) {
-      ForEach(props.children.indices, id: \.self) { index in
-        renderTabItem(at: index)
-      }
-      .measureView { size in
-        onLayout(size)
-      }
-    }
+    tabContent
 #if !os(tvOS) && !os(macOS) && !os(visionOS)
     .onTabItemEvent { index, isLongPress in
       let item = props.filteredItems[safe: index]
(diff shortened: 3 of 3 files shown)


packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
```diff
@@ -0,0 +1,69 @@
...
+            .customizationID(tabData.key)
+            .customizationBehavior(.disabled, for: .sidebar, .tabBar)
+            .accessibilityIdentifier(tabData.testID ?? "")
+          }
+        }
+      }
+    }
+    .measureView { size in
+      onLayout(size)
+    }
```
okwasniewski, line 65: Can you check if measureView returns the same values? Before we applied it to the ForEach, now it's applied to the TabView. I'm afraid this can cause issues and return different values

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
```diff
@@ -0,0 +1,69 @@
...
+                    updateTabBarAppearance(props, tabBar)
+                  #endif
+                  #if os(iOS)
+                    if index >= 4 {
+                      if props.selectedPage != tabData.key {
+                        onSelect(tabData.key)
+                      }
+                    }
+                  #endif
+                }
```
okwasniewski, line 46: I think we should pull this out to some reusable view modifier so it can be shared

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
```diff
@@ -0,0 +1,69 @@
...
+            } label: {
+              TabItem(
+                title: tabData.title,
+                icon: icon,
+                sfSymbol: tabData.sfSymbol,
+                labeled: props.labeled
+              )
+            }
+            //.badge(tabData.badge)
```
okwasniewski, line 55: Why it's commented out?

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
```diff
@@ -0,0 +1,69 @@
...
+
+  #if os(macOS)
+    var tabBar: NSTabView?
+  #else
+    var tabBar: UITabBar?
+  #endif
+
+  var onLayout: (_ size: CGSize) -> Void
+  var onSelect: (_ key: String) -> Void
+  var updateTabBarAppearance: (_ props: TabViewProps, _ tabBar: UITabBar?) -> Void
```
okwasniewski, line 16: Can we add a shared protocol that those two tabviews implement?

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
```diff
@@ -0,0 +1,69 @@
+import SwiftUI
+
+@available(iOS 18, macOS 15, visionOS 2, tvOS 18, *)
+struct NewTabView: View {
+  @ObservedObject var props: TabViewProps
+  @AppStorage("sidebarCustomizations") var tabViewCustomization: TabViewCustomization
```
okwasniewski, line 6: Can we remove tabViewCustomization for now? Let's add it in a separate PR

packages/react-native-bottom-tabs/ios/TabView/LegacyTabView.swift
```diff
@@ -0,0 +1,64 @@
...
+
+#if os(macOS)
+  var tabBar: NSTabView?
+#else
+  var tabBar: UITabBar?
+#endif
+  
+  var onLayout: (_ size: CGSize) -> Void
+  var onSelect: (_ key: String) -> Void
+  var updateTabBarAppearance: (_ props: TabViewProps, _ tabBar: UITabBar?) -> Void
```
okwasniewski, line 14: Looks like this is not working properly as on first mount the text is really small and this method is called with a `nil` tabBar.

okwasniewski: Also we probably dont need to pass the whole updateTabBarAppearance, we can make the New/LegacyTabView call a method like `onChildAppear` and then we dont need to retain the tab view here.

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
47PADO47: hi, when testing i haven't had any problems with measureView applied to TabView, and found that it would always return one value/the same value as if measureView was applied to Tab child. also i tried moving it up to the ForEach, but it gave this error `Referencing instance method 'measureView(onLayout:)' on 'ForEach' requires that 'some TabContent<String>' conform to 'View'` which i did not know how to resolve. Let me know if you think we should rework it and apply to the ForEach anyway, otherwise i'd leave it like this since it seems like nothing changes

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
47PADO47: done, let me know your thoughts on `TabAppearModifier`

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
47PADO47: done, i called it `public protocol AnyTabView`

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
47PADO47: done

packages/react-native-bottom-tabs/ios/TabView/LegacyTabView.swift
47PADO47: done, let me know if the implementation is correct

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
```diff
@@ -0,0 +1,55 @@
...
+  @ViewBuilder
+  var body: some View {
+    TabView(selection: $props.selectedPage) {
+      ForEach(props.children.indices, id: \.self) { index in
+        if let tabData = props.items[safe: index] {
+          let isFocused = props.selectedPage == tabData.key
+
+          if !tabData.hidden || isFocused {
+            let icon = props.icons[index]
+            let role: TabRole? = nil
```
hennessyevan, line 20: We can look for a `String` of `search` (To make this backwards-compatible <18.0) and then convert that into a `TabRole.search`


```suggestion
            var role: TabRole? {
              if tabData.tabRole == "search" {
                return TabRole.search
              }
              
              return nil
            }
```

hennessyevan: Of course we'd have to add the `tabRole` prop to `TabViewProvider` as well
```diff
diff --git a/packages/react-native-bottom-tabs/ios/TabViewProvider.swift b/packages/react-native-bottom-tabs/ios/TabViewProvider.swift
index 135c9a9..63669b5 100644
--- a/packages/react-native-bottom-tabs/ios/TabViewProvider.swift
+++ b/packages/react-native-bottom-tabs/ios/TabViewProvider.swift
@@ -13,7 +13,8 @@ public final class TabInfo: NSObject {
   public let activeTintColor: PlatformColor?
   public let hidden: Bool
   public let testID: String?
-
+  public let tabRole: String?
+  
   public init(
     key: String,
     title: String,
@@ -21,7 +22,8 @@ public final class TabInfo: NSObject {
     sfSymbol: String,
     activeTintColor: PlatformColor?,
     hidden: Bool,
-    testID: String?
+    testID: String?,
+    tabRole: String?
   ) {
     self.key = key
     self.title = title
@@ -30,6 +32,7 @@ public final class TabInfo: NSObject {
     self.activeTintColor = activeTintColor
     self.hidden = hidden
     self.testID = testID
+    self.tabRole = tabRole
     super.init()
   }
 }
@@ -267,7 +270,8 @@ public final class TabInfo: NSObject {
             sfSymbol: itemDict["sfSymbol"] as? String ?? "",
             activeTintColor: RCTConvert.uiColor(itemDict["activeTintColor"] as? NSNumber),
             hidden: itemDict["hidden"] as? Bool ?? false,
-            testID: itemDict["testID"] as? String ?? ""
+            testID: itemDict["testID"] as? String ?? "",
+            tabRole: itemDict["tabRole"] as? String ?? ""
           )
         )
       }
```

hennessyevan (commented):
Make `TabRole` work in the PR

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
47PADO47: makes sense to me, however i don't know if setting tabRole to a String? would be the right move, maybe using an Enum would make it more typesafe and future proof. Also, similarly to tabBarCustomization, it's probably better to add it in a follow up pr after this one is merged.

hennessyevan: Yeah `String?` isn't super ideal but the `TabRole` enum is only available in iOS18+ which would mean splitting into `NewTabViewProvider` and `LegacyTabViewProvider` as well as the split you already have.

There is currently only one role: .search but I think we could do `TabRole(rawValue: tabData.tabRole)` though which would still future proof it in case of future roles.

For sure can be done in a followup PR, it would obviously build upon yours and I couldn't figure out how to get GitHub to stack PRs

47PADO47: you're right, that's probably the easiest way atm, and i guess we could implement the enum on the TS types (`type TabRole = 'search';`) so typescript users will have a typesafe way of adding a tab role. As for pr stacking i don't know either, we might have to wait for this one to be merged first

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
hennessyevan: I think that's because it's supposed to be `Text?` not `String?`

This compiled for me
```swift
  .badge(tabData.badge.isEmpty ? nil : Text(tabData.badge))
```

47PADO47: thanks for pointing it out. when developing i only saw and looked into the one below, which didn't specify text nor string

https://developer.apple.com/documentation/swiftui/tabcontent/badge(_:)

currently i can't test your implementation, however if you say it works with Text then this might be the solution. can you confirm the problem is gone on both ios 18.5 and ios 26 simulators?

okwasniewski: @47PADO47 I've created a follow up PR addressing the behavior of the empty badge prop: https://github.com/callstackincubator/react-native-bottom-tabs/pull/378

It should unblock this PR!

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
okwasniewski: I would remove passing role in this PR completely and create a separate PR implementing it. For PR stacking you can create a new branch off this PR and create a new PR pointing to this branch. 

Making it a string sounds good to me, we can use TypeScript to enforce users to pass `search` and if in the future Apple adds more roles we will be ready.

okwasniewski (commented):
We are getting there! Thanks for applying the fixes. 

Sorry for taking so long to review but I got sick last week 🥲 

I'm going to merge this PR soon: https://github.com/callstackincubator/react-native-bottom-tabs/pull/378

Can you rebase and implement the badge behavior properly? Afterwards we should be good to merge, also can you run `swiftlint --fix ./packages` in the root of the repo?

47PADO47:
no problem, take your time to recover.
glad we've been able to tackle the badge problem, i'll rebase and push fixes as soon as i can

packages/react-native-bottom-tabs/ios/TabViewImpl.swift
```diff
@@ -13,21 +13,37 @@ struct TabViewImpl: View {
 #else
   @Weak var tabBar: UITabBar?
 #endif
-
+  
+  @ViewBuilder
+  var tabContent: some View {
+    if #available(iOS 18, macOS 15, visionOS 2, tvOS 18, *) {
+      NewTabView(
```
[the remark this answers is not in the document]
47PADO47: correct me if i'm wrong, but i don't think it's necessary as onSelect/onLongPress events are handled by .onTabItemEvent in TabViewImpl.swift for both implementations

okwasniewski:
Thank you for working on this! Just those two fixes I found were necessary to make old tab view work as before

PR #387, opened 2025-07-12
feat: add tab roles/search role

47PADO47:
## PR Description

This pr adds support for [ios tab roles](https://developer.apple.com/documentation/swiftui/tabrole/search)

## How to test?

Run the app on a mac and go into "sf symbols" example

## Screenshots

Simulator Screenshot - iPhone 16 Pro - 2025-07-12 at 18 53 46

simulator_screenshot_C37172EB-9DD1-4D0E-84E4-F7CA1D94BDD1

### DIFF
--- a/docs/docs/docs/guides/standalone-usage.md
+++ b/docs/docs/docs/guides/standalone-usage.md
@@ -212,6 +212,7 @@ Each route in the `routes` array can have the following properties:
 - `activeTintColor`: Custom active tint color for this specific tab
 - `lazy`: Whether to lazy load this tab's content
 - `freezeOnBlur`: Whether to freeze the tab's content when it's not visible
+- `role`: A value that defines the purpose of the tab
 
 ### Helper Props
 
@@ -260,3 +261,9 @@ Function to determine if a tab should be hidden.
 Function to get the test ID for a tab item.
 
 - Default: Uses `route.testID`
+
+#### `getRole`
+
+Function to get the role for a tab item.
+
+- Default: Uses `route.role`


okwasniewski:
@47PADO47 Thanks for working on this! Can you rebase with latest changes so only the changes adding tab roles are in this PR?

packages/react-navigation/src/types.ts
```diff
@@ -92,6 +92,11 @@ export type NativeBottomTabNavigationOptions = {
    */
   tabBarButtonTestID?: string;
 
+  /**
+   * role for the tab. (iOS only)
```
okwasniewski, line 96:
```suggestion
   * Role for the tab. (iOS only)
```

packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabViewImpl.kt
```diff
@@ -35,7 +36,8 @@ class RCTTabViewImpl {
               badge = if (item.hasKey("badge")) item.getString("badge") else null,
               activeTintColor = if (item.hasKey("activeTintColor")) item.getInt("activeTintColor") else null,
               hidden = if (item.hasKey("hidden")) item.getBoolean("hidden") else false,
-              testID = item.getString("testID")
+              testID = item.getString("testID"),
+              role = item.getString("role"),
```
okwasniewski, line 40: There is no need to add it to Android implementation

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
```diff
@@ -17,6 +17,13 @@ struct NewTabView: AnyTabView {
 
           if !tabData.hidden || isFocused {
             let icon = props.icons[index]
+            var role: TabRole? {
+              if tabData.role == "search" {
+                return .search
+              }
+
+              return nil
+            }
```
okwasniewski, line 26: We should encapsulate this conversion into an enum same as it was done here: https://github.com/callstackincubator/react-native-bottom-tabs/blob/c1a0dd7faf88d7b6aeaa5c321cd6c4be7ba889af/packages/react-native-bottom-tabs/ios/TabViewProps.swift#L3-L24 

So its nicely separated

okwasniewski (commented):
Thanks for working on this! 

Can you also update documentation for both React Navigation usage and standalone usage?

47PADO47:
@okwasniewski i think everything's done, what do you say?

PR #393, opened 2025-07-25
Apply typeface and size individually in android text appearance.

joarkosberg:
When using a custom font with bottom tabs today on Android, another font size is also set by default. With the biggest difference from the standard one being that it scales with the system font scale, which the default one does not. 

## PR Description

**Improvement**: This splits the application of text appearance, so that size and typeface(family and weight) is applied individually. So that when setting one or the other, only that is applied. So one could use a custom font family, without getting the font scaling which is usually not wanted with tab bars.

## How to test?

On Android apply `fontFamily`+`fontWeight` and `fontSize` individually in `tabLabelStyle`, and see they can be used individually.

## Screenshots

Before and after examples with bigger system font scale on Android
| Code | Before | After |
|--------|--------|--------|
|Screenshot 2025-07-25 at 09 14 29 | Screenshot 2025-07-25 at 09 03 36 | Screenshot 2025-07-25 at 08 55 09 |
| Screenshot 2025-07-25 at 09 13 50| Screenshot 2025-07-25 at 08 56 25 | (same as before) Screenshot 2025-07-25 at 08 56 25 |

### DIFF
--- a/.changeset/bright-windows-sip.md
+++ b/.changeset/bright-windows-sip.md
@@ -1,0 +1,5 @@
+---
+"react-native-bottom-tabs": patch
+---
+
+fix: apply typeface and size individually in android text appearance.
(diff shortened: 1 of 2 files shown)


okwasniewski (approved):
Looks good, thanks for working on this

PR #415, opened 2025-08-27
fix: improve subview foreach, use view as identifier

okwasniewski:
## PR Description

This PR should improve the subview foreach on iOS using the view as identifier not the index.


thymikee (approved):
👍🏼

PR #426, opened 2025-09-12
chore(android): enable flexible page sizes

seyedmostafahasani:
## PR Description

This PR adds support for 16 KB memory page size on Android 15+ devices.
To achieve this, the build.gradle file was updated to enable flexible page size support in the native build process by passing the following argument to CMake:
```
externalNativeBuild {
  cmake {
    arguments "-DANDROID_SUPPORT_FLEXIBLE_PAGE_SIZES=ON"
  }
}
```
This ensures the React Native app can run correctly on newer Android devices that ship with a 16 KB page size by default.

## How to test?

1. Build and install the app on an Android 15 emulator or device configured with a 16 KB page size.
2. Verify that the app launches and runs without native crashes related to memory mapping or page alignment.
3. Optionally, test on devices with 4 KB page size to confirm backward compatibility.

## Screenshots

No UI changes.


okwasniewski:
Hey, thanks for your PR.

Do you think it's necessary to add this? Shouldn't it be handled by React Native?

seyedmostafahasani:
Hey, Anytime!
I think it will take a while until React Native implements it, so we can handle it for now since some developers might need it. WDYT?

thiagobrez:
Seems like this is not needed anymore, since React Native added full support for 16KB pages on v0.77.

Also, we do not compile our own libraries with CMake, so this shouldn't be necessary.

Closing for now, let us know if this is still causing issues 👌🏻

PR #494, opened 2025-12-16
feat: Add `layoutDirection` prop to `TabView` for RTL support

ahmedawaad1804:
##RTL Support 

## PR Description

This PR adds full Right-to-Left (RTL) support to the Bottom Tabs implementation.
The tab layout direction now correctly follows the provided layout direction, ensuring proper ordering, selection behavior, and visual consistency for RTL languages (e.g. Arabic).

## How to test?

1- Run app and go to last Example (Four Tabs - RTL)
2- it will automatically force RTL 
* will return to LTR when unmount

## Screenshots

https://github.com/user-attachments/assets/0f08612b-89e5-43be-8ba1-b64e9b4f2289

### DIFF
--- a/packages/react-native-bottom-tabs/src/TabView.tsx
+++ b/packages/react-native-bottom-tabs/src/TabView.tsx
@@ -201,6 +201,12 @@ interface Props<Route extends BaseRoute> {
    * @platform ios
    */
   renderBottomAccessoryView?: BottomAccessoryViewProps['renderBottomAccessoryView'];
+  /**
+   * The direction of the layout. (iOS only)
+   * @platform ios
+   * @default 'leftToRight'
+   */
+  layoutDirection?: 'leftToRight' | 'rightToLeft';
 }
 
 const ANDROID_MAX_TABS = 100;
@@ -239,6 +245,7 @@ const TabView = <Route extends BaseRoute>({
   tabBarStyle,
   tabLabelStyle,
   renderBottomAccessoryView,
+  layoutDirection = 'leftToRight',
   ...props
 }: Props<Route>) => {
   // @ts-ignore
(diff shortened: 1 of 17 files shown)


packages/react-native-bottom-tabs/src/TabView.tsx
```diff
@@ -201,6 +201,12 @@ interface Props<Route extends BaseRoute> {
    * @platform ios
    */
   renderBottomAccessoryView?: BottomAccessoryViewProps['renderBottomAccessoryView'];
+  /**
+   * The direction of the layout. (iOS only)
+   * @platform ios
+   * @default 'leftToRight'
+   */
+  layoutDirection?: 'leftToRight' | 'rightToLeft';
```
okwasniewski: We should allow for three options:

`layoutDirection: 'ltr' / 'rtl' / 'locale'`

Using `ltr` and `rtl` as already established patterns. `locale` is the default.

We also need to support Android if we want to roll out this feature.

okwasniewski (changes requested):
Hey,

Thanks for working on this! 

We will need to have support for this prop on both platforms. It should be pretty straight forward.

Another thing is that we need to expose this prop to React Navigation integration in `packages/react-navigation`

packages/react-native-bottom-tabs/src/TabView.tsx
ahmedawaad1804: Done

ahmedawaad1804:
@okwasniewski 
Added to android with default value locale to follow system layout

ahmedawaad1804:
> Hey,
> 
> Thanks for working on this!
> 
> We will need to have support for this prop on both platforms. It should be pretty straight forward.
> 
> Another thing is that we need to expose this prop to React Navigation integration in `packages/react-navigation`

Done

okwasniewski (approved):
Looks good, thank you!

Nader-CS:
plesae can you release a version ? @okwasniewski

PR #508, opened 2026-03-27
fix: add tvOS availability annotations for iOS 26 APIs

oscnord:
## Summary

Building for tvOS fails with compilation errors because several iOS 26 APIs lack tvOS availability annotations.

### Errors fixed:
- `BottomAccessoryProvider.swift` — `TabViewBottomAccessoryPlacement` needs `@available(tvOS 26.0)`
- `NewTabView.swift` — `.tabViewBottomAccessory` modifier is unavailable on tvOS, excluded via `#if !os(tvOS)`
- `TabViewProps.swift` — `TabBarMinimizeBehavior` enum values (`.never`, `.onScrollUp`, `.onScrollDown`) are unavailable on tvOS, falls back to `.automatic`
- `TabViewImpl.swift` — `tabBarMinimizeBehavior` modifier needs `tvOS 26.0` in `#available` check

### Changes:
- Added `tvOS 26.0` to `@available` / `#available` annotations where the API exists on tvOS
- Excluded `tabViewBottomAccessory` and `BottomAccessoryRepresentableView` from tvOS via `#if` (API doesn't exist on tvOS at all)
- Removed `tvOS 26.0` from the `#available(iOS 26.0, tvOS 26.0, visionOS 3.0, *)` check in `NewTabView.swift` — the original included tvOS in the runtime check, but `tabViewBottomAccessory` doesn't exist on tvOS even in 26.0, so the check would pass but compilation would fail. The block is now excluded at compile time via `#if os(tvOS)` instead.
- Fall back to `.automatic` for `MinimizeBehavior.convert()` on tvOS (same as macOS)

### Tested:
- tvOS build compiles successfully with these changes


thiagobrez:
Thank you @oscnord! This looks very reasonable. I will be testing soon.

thiagobrez (approved):
LGTM, tested on a fresh new RN tvOS app. Thank you!

thiagobrez:
Released in `1.3.0` 🎉

Issue #439, opened 2025-10-01
[iOS 26] tabBarInactiveTintColor does not apply to tab bar icons (only labels)


beau-house:
### Before submitting a new issue

- [x] I tested using the latest version of the library, as the bug might be already fixed.
- [x] I tested using a [supported version](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md) of react native.
- [x] I checked for possible duplicate issues, with possible answers.

### Bug summary

The `tabBarInactiveTintColor` prop does not apply to inactive tab icons, while inactive labels correctly show the specified colour.

### Expected behaviour
Both icons and labels should respect `tabBarInactiveTintColor`.

### Suspected cause
SwiftUI's `.tintColor()` modifier in `TabViewImpl.swift:76` applies the active color globally, overriding the inactive `iconColor` set in `UITabBarItemAppearance (line 175)`. Labels use `titleTextAttributes`, which aren't affected by the SwiftUI tint modifier.

https://github.com/user-attachments/assets/b08161b9-5abb-4396-9d94-8cb9040b0b25

### Library version

0.11.2

### Environment info

```shell
System:
  OS: macOS 26.0.1
  CPU: (10) arm64 Apple M1 Max
  Memory: 114.78 MB / 32.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 22.12.0
    path: ~/.nvm/versions/node/v22.12.0/bin/node
  Yarn:
    version: 4.6.0
    path: ~/.nvm/versions/node/v22.12.0/bin/yarn
  npm:
    version: 11.5.2
    path: ~/.nvm/versions/node/v22.12.0/bin/npm
  Watchman:
    version: 2025.09.15.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.16.2
    path: /usr/local/lib/ruby/gems/3.3.0/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.0
      - iOS 26.0
      - macOS 26.0
      - tvOS 26.0
      - visionOS 26.0
      - watchOS 26.0
  Android SDK: Not Found
IDEs:
  Android Studio: 2022.3 AI-223.8836.35.2231.10406996
  Xcode:
    version: 26.0.1/17A400
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.16
    path: /usr/bin/javac
[...]

PR #509, opened 2026-03-30
fix(ios): preserve Liquid Glass on iOS 26 by skipping custom UITabBarAppearance

deepak0x:
## Problem

On iOS 26, the system provides a Liquid Glass material for `UITabBar` automatically. However, `updateTabBarAppearance()` in `TabViewImpl.swift` calls `configureStandardAppearance()`, which creates a custom `UITabBarAppearance` with explicit background and item styling. Apple's WWDC25 session 284 ("Build a UIKit app with the new design") says:

> "Remove any background customization from your navigation and toolbars. Using UIBarAppearance or backgroundColor interferes with the glass appearance."

So any app using `react-native-bottom-tabs` on iOS 26 gets an opaque tab bar instead of Liquid Glass, even with `UIDesignRequiresCompatibility = false` in Info.plist.

## Fix

Add an early return in `updateTabBarAppearance()` on iOS 26+ when:
- No explicit `barTintColor` is set (the app wants the system default)
- `scrollEdgeAppearance` is not `"opaque"` (the app is not requesting an opaque bar)

In this case, we skip `configureStandardAppearance()` and let iOS render Liquid Glass. The inactive tint color is still applied via `tabBar.unselectedItemTintColor` for apps that set `tabBarInactiveTintColor`.

If the app sets a custom `barTintColor` or requests an opaque appearance, the existing code path runs as before. On iOS 25 and below, the early return is guarded by `#available(iOS 26.0, *)` so the existing code path is completely unchanged.

## Testing

Tested with the Rocket.Chat React Native app connected to a live server. Liquid Glass renders on the tab bar after the fix. Without it, the tab bar is opaque.

Related to #439


deepak0x:
hey @maintainers pls check this pr 
Thanks ;)

thiagobrez:
Hey @deepak0x , thanks for the PR! Would you mind sharing some screenshots and relevant snippets / prop configuration to achieve them 🙏🏻 

Thank you

deepak0x:
Hey @thiagobrez, thanks for the quick look! Here are the prop configurations corresponding to the three branches the fix touches. Screenshots attached below.

### 1. Liquid Glass (new behavior on iOS 26)
No `barTintColor`, no opaque request — iOS renders its native Liquid Glass material.

```jsx
<TabView
  navigationState={{ index, routes }}
  onIndexChange={setIndex}
  renderScene={renderScene}
  // no barTintColor
  // no scrollEdgeAppearance="opaque"
  tabBarInactiveTintColor="#8E8E93" // still applied via unselectedItemTintColor
/>
```

### 2. Custom opaque bar (existing path, unchanged)
Setting `barTintColor` keeps the original `configureStandardAppearance()` code path so the custom color is honored.

```jsx
<TabView
  navigationState={{ index, routes }}
  onIndexChange={setIndex}
  renderScene={renderScene}
  barTintColor="#FFFFFF"
/>
```

### 3. Explicit opaque request (existing path, unchanged)
Setting `scrollEdgeAppearance="opaque"` also runs the original path, so apps that intentionally want an opaque bar on iOS 26 still get one.

```jsx
<TabView
  navigationState={{ index, routes }}
  onIndexChange={setIndex}
  renderScene={renderScene}
  scrollEdgeAppearance="opaque"
/>
```

On iOS 25 and below, all three configurations behave exactly as before — the early return is guarded by `#available(iOS 26.0, *)`.

Screenshots below 
WhatsApp Image 2026-05-04 at 5 16 55 AM

before this fix it was a  opaque tab bar i.e. configuration two...!

thiagobrez:
@deepak0x Appreciate it! I'll try to find the time to look into this again this week.

FYI: I think you forgot to attach the screenshots 😆

thiagobrez:
@deepak0x I don't seem to be able to reproduce on iOS 26.4 🤔 which version are you using?

deepak0x:
Hey @thiagobrez — sorry for the long silence, and thanks for taking the time to test this.

To answer your question: I originally reproduced this on **iOS 26.0**, back when I opened the PR in March.

I went back and re-tested on **iOS 26.1** (iPhone 17 Pro simulator, Xcode 26.1.1) — building the example app from `main` *without* this PR's change — and I can no longer reproduce it there either. The tab bar renders the proper Liquid Glass material and the inactive tint is respected. That lines up with your 26.4 result, so it looks like the behaviour changed on Apple's side somewhere after 26.0.

(That's also why I didn't re-attach screenshots — on 26.1 the unpatched build already looks correct, so there's no "before" state left to capture.)

I don't want to overclaim that the issue is fully gone, though — #439 still has reports against 26.2, and I've only checked the 26.1 simulator, not a physical device or every point release.

So I'm happy to go whichever way you prefer:
- close this PR if you consider it obsolete for current iOS versions, or
- keep it as a safeguard for 26.0 and any later versions where it resurfaces.

Let me know what you'd like — and apologies again for the delay.

thiagobrez:
No worries @deepak0x, we all have stuff to do :)

Thanks for the double check. I'm investigating #439 as well, and unfortunately there are many undocumented changes on Apple's side to the tab bar and tint behavior between 26.0 - 26.5, which makes it harder to debug.

Since this one seems to be resolved, I will close this for simplicity. I will continue on the other issue.

Thank you!

PR #519, opened 2026-05-15
fix(iOS): janky tab switch between colored and uncolored tabs

thiagobrez:
## PR Description

When switching between a tint-colored tab and a standard tab, the native component was remounting and causing the tab switch to cut mid-animation, feeling janky.

This fix makes it so the Tab keeps mounted and switch stays smooth.


thiagobrez:
Released in `1.3.0` 🎉

PR #522, opened 2026-05-19
feat(iOS): experimental baked tint colors

thiagobrez:
## PR Description

Apple has been shipping many undocumented changes between iOS versions >26. iOS is now overriding the tab inactive tint colors forcefully.

Due to this, on #527 inactive tint colors were disabled for iOS 26+.

This PR adds an experimental prop `experimental_bakedTintColors` that changes the way the icons and labels are rendered, baking (rasterizing) the label within the icon image. Rhis seems to be the only workaround to making inactive tint colors work.

However, this has many drawbacks, such as SFSymbol icon sizes being different on label width, badges being positioned far away from the icon depending on the label width, and possibly breaking accessibility since the label will be baked inside the icon image.

Use with caution, only if you prioritize tint color consistency between platforms.

## How to test?

- Tested on iOS 18.6, 26.0.1, 26.1, 26.2, 26.3, 26.4.1 and 26.5
- Tested with custom icons and SFSymbols
- Tested with remote icons

### Before

https://github.com/user-attachments/assets/298cb930-8713-4ffc-b3a4-f9c1a5b5a418

### After

https://github.com/user-attachments/assets/c143cb02-fab1-4fc7-9a44-792fb0c704d7

### DIFF
--- a/.changeset/grumpy-otters-report.md
+++ b/.changeset/grumpy-otters-report.md
@@ -1,0 +1,5 @@
+---
+'react-native-bottom-tabs': patch
+---
+
+Disable tabBarInactiveTintColor on iOS >= 26
--- a/.changeset/large-hats-teach.md
+++ b/.changeset/large-hats-teach.md
@@ -2,4 +2,4 @@
 'react-native-bottom-tabs': patch
 ---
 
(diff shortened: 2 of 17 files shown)


thiagobrez:
Released in `1.3.0` 🎉

Issue #76, opened 2024-10-27
Implement `disappearOnScroll` 


okwasniewski:
Goal of this issue is to implement `disappearOnScroll` for Android and check if it's possible for iOS.

anudit:
@okwasniewski Looking forward to this

cesargdm:
Expo implements it with https://docs.expo.dev/router/advanced/tabs/#hiding-a-tab

okwasniewski:
Hey @cesargdm, `options={{ href: null }}` is used for hiding a screen not the tab bar. 

The screen hiding is also supported using `tabBarItemHidden`, supporting `href: null` is possible but would require creating an expo specific package with the navigator (they add it on top of React Navigation tabs too).

rochajulian:
Any update in this? The way that I used to do it on Expo is as follows:
```
<Tabs.Screen
    name="index"
    options={{
        href: null,
        tabBarStyle: {
            display: "none"
        }
    }}
/>
```      
  This would effectively hide the screen from the tab bar AND the tab bar on that screen.

PR #524, opened 2026-05-29
fix(TabView): respect user-supplied tabBarHidden

### LINKED ISSUE #521: `tabBarHidden` navigator prop is silently ignored — hardcoded in JS wrapper
## Summary

`tabBarHidden` is documented and wired up natively (Swift `.hideTabBar(props.tabBarHidden)`, addressed in #76), but the JS wrapper in `packages/react-native-bottom-tabs/src/TabView.tsx` hardcodes it on the inner `NativeTabView`, so any user-supplied value is silently overwritten before it reaches native.

## Repro

```tsx
import TabView from 'react-native-bottom-tabs';

// — or with the React Navigation adapter —
// <Tabs.Navigator tabBarHidden={shouldHide}>

<TabView
  navigationState={{ index, routes }}
  renderScene={renderScene}
  onIndexChange={setIndex}
  tabBarHidden={shouldHide}   // ← never reaches native
/>
```

Expected: the native tab bar is hidden via SwiftUI's `.hideTabBar(true)` when `shouldHide === true`.

Actual: tab bar stays visible regardless of the value passed. The user-supplied prop is shadowed inside `TabView.tsx` by:

```tsx
// packages/react-native-bottom-tabs/src/TabView.tsx (~line 395 on `main`)
<NativeTabView
  {...props}
  ...
  tabBarHidden={!!renderCustomTabBar}   // ← overwrites props.tabBarHidden
  ...
/>
```

The compiled output in `lib/module/TabView.js` has the same shape:

```js
tabBarHidden: !!renderCustomTabBar,
```

Because the line is *after* `{...props}`, JSX prop precedence wins and the user's `tabBarHidden` is discarded.

## Why this matters on iOS 26

External native-module workarounds that wrote to `UITabBar.isHidden` from a `UIWindowScene` traversal worked on iOS ≤ 18 but stopped working on iOS 26, because the SwiftUI host now reconciles the bar's `isHidden` from `props.tabBarHidden` on every appearance pass (`TabViewImpl.swift` `updateTabBarAppearance`). With the JS wrapper hardcoding `tabBarHidden = !!renderCustomTabBar`, that prop is always `false` for non-custom-tab-bar consumers — so any external attempt to hide is clobbered.

Net effect: there is currently **no working path** for an app to per-screen hide the native tab bar on iOS 26 without forking or patching the library.

## Proposed fix

Honor a user-supplied `tabBarHidden` when set; fall back to the existing "hide when custom tab bar is rendered" behavior otherwise.

```diff
   <NativeTabView
     {...props}
     {...tabLabelStyle}
     style={styles.fullWidth}
     items={items}
     // When rendering a custom tab bar, icons can be React elements, which will not be properly resolved.
     icons={renderCustomTabBar ? undefined : resolvedIconAssets}
     selectedPage={focusedKey}
-    tabBarHidden={!!renderCustomTabBar}
+    tabBarHidden={props.tabBarHidden ?? !!renderCustomTabBar}
```

That preserves backwards compatibility (default behavior is unchanged for any consumer that doesn't pass `tabBarHidden`) and unblocks the documented use case.

Happy to open a PR.

## Environment

- `react-native-bottom-tabs` 1.1.0 (also present on `main` as of this filing)
- `@bottom-tabs/react-navigation` 1.0.x
- iOS 26 SDK, real device (iPhone 15 Pro Max)
- React Native 0.81.x, Expo SDK 54

spokodev:
Closes #521.

`tabBarHidden` is wired natively (SwiftUI `.hideTabBar(props.tabBarHidden)`, landed in #76) but the JS wrapper in `packages/react-native-bottom-tabs/src/TabView.tsx` hardcoded `tabBarHidden={!!renderCustomTabBar}` after the `{...props}` spread, so any caller-supplied value was overwritten before it reached `NativeTabView`. The prop was also absent from the `Props<Route>` interface so it never surfaced at the type level either.

This PR:
- adds `tabBarHidden?: boolean` to the `TabView` `Props<Route>` interface with a JSDoc note that explains the default and the iOS 26+ use case from the issue
- destructures it from the function parameters alongside the other named props
- keeps the existing `!!renderCustomTabBar` behavior as the fallback when the caller does not pass a value (`tabBarHidden ?? !!renderCustomTabBar`)

Effect:
- callers can now set `<TabView ... tabBarHidden={true} />` to hide the native bar without rendering a custom `tabBar`, which is what the issue needs on iOS 26 where the `UITabBar.isHidden` workarounds from outside React Native no longer reach the SwiftUI-backed bar
- callers that do not pass `tabBarHidden` get the same behavior as before (hidden iff `tabBar` is provided)
- TypeScript users see the prop in IntelliSense and on the public type

Verification:
- `tsc -b` clean on the package
- `eslint packages/react-native-bottom-tabs/src/TabView.tsx` reports only the pre-existing `no-shadow` warning on `loaded` (line 293), unrelated to this change
- existing jest setup in this repo is `it.todo('write a test')` and `yarn test` fails on `main` due to Flow type syntax in `react-native/jest/setup.js`, so a behavioural test would require fixing the jest config first which is outside the scope of this PR
- changeset added: `.changeset/respect-user-tab-bar-hidden.md` (`react-native-bottom-tabs: patch`)

Out of scope:
- the React Navigation adapter package (`@bottom-tabs/react-navigation`) does not reference `tabBarHidden`, so the `<Tabs.Navigator tabBarHidden={shouldHide}>` mention in the issue would need a separate change in that package's screen-options forwarding

### DIFF
--- a/.changeset/respect-user-tab-bar-hidden.md
+++ b/.changeset/respect-user-tab-bar-hidden.md
@@ -1,0 +1,5 @@
+---
+'react-native-bottom-tabs': patch
+---
+
+Respect user-supplied `tabBarHidden` on `TabView`. The prop was wired natively but the JS wrapper hardcoded `tabBarHidden={!!renderCustomTabBar}` after the `{...props}` spread, so any user value was silently overwritten. The prop is now declared on the public `TabView` API and only falls back to the custom-tab-bar default when the caller omits it, which lets iOS 26+ users hide the SwiftUI-backed tab bar from JS again.
--- a/.changeset/respect-user-tab-bar-hidden.md
+++ b/.changeset/respect-user-tab-bar-hidden.md
@@ -2,4 +2,4 @@
 'react-native-bottom-tabs': patch
 ---
 
-Respect user-supplied `tabBarHidden` on `TabView`. The prop was wired natively but the JS wrapper hardcoded `tabBarHidden={!!renderCustomTabBar}` after the `{...props}` spread, so any user value was silently overwritten. The prop is now declared on the public `TabView` API and only falls back to the custom-tab-bar default when the caller omits it, which lets iOS 26+ users hide the SwiftUI-backed tab bar from JS again.
+Respect user-supplied `tabBarHidden` on `TabView`.
--- a/packages/react-native-bottom-tabs/src/TabView.tsx
+++ b/packages/react-native-bottom-tabs/src/TabView.tsx
@@ -212,6 +212,14 @@ interface Props<Route extends BaseRoute> {
    * @default 'locale'
    */
   layoutDirection?: LayoutDirection;
+  /**
+   * Whether to hide the native tab bar. Defaults to `true` when a custom
+   * `tabBar` is provided (so the native bar does not stack on top of the
+   * custom one), otherwise `false`. Set explicitly to override the default,
+   * which is required on iOS 26+ where `UITabBar.isHidden` workarounds from
+   * outside React Native no longer reach the SwiftUI-backed tab bar.
+   */
+  tabBarHidden?: boolean;
 }
 
 const ANDROID_MAX_TABS = 100;
@@ -247,6 +255,7 @@ const TabView = <Route extends BaseRoute>({
   labeled = Platform.OS !== 'android' ? true : undefined,
   getFreezeOnBlur = ({ route }: { route: Route }) => route.freezeOnBlur,
   tabBar: renderCustomTabBar,
+  tabBarHidden,
   tabBarStyle,
   tabLabelStyle,
   renderBottomAccessoryView,
@@ -404,7 +413,7 @@ const TabView = <Route extends BaseRoute>({
         // When rendering a custom tab bar, icons can be React elements, which will not be properly resolved.
         icons={renderCustomTabBar ? undefined : resolvedIconAssets}
         selectedPage={focusedKey}
-        tabBarHidden={!!renderCustomTabBar}
+        tabBarHidden={tabBarHidden ?? !!renderCustomTabBar}
         onTabLongPress={handleTabLongPress}
         onPageSelected={handlePageSelected}
         onTabBarMeasured={handleTabBarMeasured}
--- a/packages/react-native-bottom-tabs/src/TabView.tsx
+++ b/packages/react-native-bottom-tabs/src/TabView.tsx
@@ -418,7 +418,7 @@ const TabView = <Route extends BaseRoute>({
         // When rendering a custom tab bar, icons can be React elements, which will not be properly resolved.
         icons={renderCustomTabBar ? undefined : resolvedIconAssets}
         selectedPage={focusedKey}
-        tabBarHidden={tabBarHidden ?? !!renderCustomTabBar}
+        tabBarHidden={tabBarHidden}
         onTabLongPress={handleTabLongPress}
         onPageSelected={handlePageSelected}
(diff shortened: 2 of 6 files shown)


packages/react-native-bottom-tabs/src/TabView.tsx
```diff
@@ -247,6 +255,7 @@ const TabView = <Route extends BaseRoute>({
   labeled = Platform.OS !== 'android' ? true : undefined,
   getFreezeOnBlur = ({ route }: { route: Route }) => route.freezeOnBlur,
   tabBar: renderCustomTabBar,
+  tabBarHidden,
```
thiagobrez, line 262: Please add the default `false` value 🙏🏻 

```suggestion
  tabBarHidden = false,
```

packages/react-native-bottom-tabs/src/TabView.tsx
```diff
@@ -212,6 +212,14 @@ interface Props<Route extends BaseRoute> {
    * @default 'locale'
    */
   layoutDirection?: LayoutDirection;
+  /**
+   * Whether to hide the native tab bar. Defaults to `true` when a custom
+   * `tabBar` is provided (so the native bar does not stack on top of the
+   * custom one), otherwise `false`. Set explicitly to override the default,
+   * which is required on iOS 26+ where `UITabBar.isHidden` workarounds from
+   * outside React Native no longer reach the SwiftUI-backed tab bar.
+   */
```
thiagobrez, line 225: No need for such a big wall of text, let's keep it simple

```suggestion
  /**
   * Whether to hide the native tab bar.
   */
```

.changeset/respect-user-tab-bar-hidden.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-native-bottom-tabs': patch
+---
+
+Respect user-supplied `tabBarHidden` on `TabView`. The prop was wired natively but the JS wrapper hardcoded `tabBarHidden={!!renderCustomTabBar}` after the `{...props}` spread, so any user value was silently overwritten. The prop is now declared on the public `TabView` API and only falls back to the custom-tab-bar default when the caller omits it, which lets iOS 26+ users hide the SwiftUI-backed tab bar from JS again.
```
thiagobrez: Same here, we don't need the whole explanation on the changelogs

```suggestion
Respect user-supplied `tabBarHidden` on `TabView`.
```

thiagobrez:
Thanks @spokodev! Left some easy comments. If you can fix them and mark the PR as ready, we can merge soon!

spokodev:
Pushed a small follow-up on line 421 -- the `?? !!renderCustomTabBar` fallback is unreachable with the new `tabBarHidden = false` default, so removed it. One trade-off worth flagging: callers who previously relied on the auto-hide-when-renderCustomTabBar behavior now need to set `tabBarHidden={true}` explicitly. The new examples you just pushed look aligned with that direction. Can revert the line 421 change if a different approach works better. Thanks!

thiagobrez:
@spokodev Hmmm good catch, thanks for spotting that. We might want to remove the `default = false` though, let me check...

thiagobrez:
@spokodev I rolled back your last change because I was wrong from the beginning adding that default false value. Thanks for catching. I also added more examples and will be merging as soon as CI pass 🎉 

btw - sorry for taking over, I'm planning to do a release today and wanted this in :)

thiagobrez (approved):

thiagobrez:
Released in `1.3.0` 🎉

PR #526, opened 2026-06-06
fix(ios): gate index>=4 appear-select on >5-tab overflow

### LINKED ISSUE #525: TabAppearModifier hijacks the 5th tab (index 4) selection when a hidden tab bar re-appears
## Describe the bug

On iOS, `TabAppearModifier` force-selects any tab whose `index >= 4` in `onAppear`:

```swift
#if os(iOS)
  if context.index >= 4, context.props.selectedPage != context.tabData.key {
    context.onSelect(context.tabData.key)
  }
#endif
```

This is intended to sync selection for tabs nested under iOS's **"More"** overflow tab, which only exists when there are **more than 5 visible tabs**. With exactly 5 tabs there is no More tab, but the 5th tab (index 4) still matches `index >= 4`. So whenever that tab's scene re-appears — e.g. when the tab bar is re-shown after a screen hid it via `.hideTabBar` (`true` → `false`) — it fires `onSelect` and **steals the selection**, producing a phantom tab switch.

It cannot be corrected from JS: `selectedPage` is a value-diffed controlled prop and there's no imperative setter, so once the native bar drifts, JS re-passing the same `selectedPage` is a no-op.

## To Reproduce

With a 5-tab native bottom tabs setup (here via `@bottom-tabs/react-navigation` + expo-router):

1. Select the 5th tab (index 4) once.
2. Navigate to a nested screen that sets `tabBarHidden = true`.
3. Press back (tab bar re-shows, `tabBarHidden = false`).

➡️ The app jumps to the 5th tab instead of staying on the tab you returned to.

## Root cause

On iOS the only code paths that emit `onPageSelected` are (a) a real user tab tap (`onTabItemEvent`) and (b) this `index >= 4` appear-handler. In the repro there is no tap, so the switch comes from (b): the 5th tab's `onAppear` fires during the tab-bar re-materialization and force-selects itself.

## Fix

Gate the force-select on actual overflow (more than 5 visible tabs, i.e. when a real "More" tab exists):

```diff
 #if os(iOS)
-  if context.index >= 4, context.props.selectedPage != context.tabData.key {
+  if context.props.filteredItems.count > 5, context.index >= 4, context.props.selectedPage != context.tabData.key {
     context.onSelect(context.tabData.key)
   }
 #endif
```

This preserves the existing behavior for >5-tab (More menu) configurations and removes the spurious selection for ≤5-tab bars.

## Versions

Reproduced on `react-native-bottom-tabs` 1.1.0 and 1.2.0 (iOS).

davidecallegaro:
## Summary

Fixes #525.

On iOS, `TabAppearModifier` force-selects any tab with `index >= 4` in `onAppear` to sync selection for tabs nested under iOS's **"More"** overflow tab. That overflow only exists when there are **more than 5 visible tabs**. With exactly 5 tabs there is no More tab, yet the 5th tab (index 4) still matches `index >= 4`, so it steals the selection whenever its scene re-appears — most visibly when the tab bar is re-shown after a screen hid it via `.hideTabBar` — causing a phantom tab switch that JS can't correct (`selectedPage` is a value-diffed controlled prop).

## Change

Gate the force-select on actual overflow (`filteredItems.count > 5`), so it only runs when a real "More" tab exists.

```diff
 #if os(iOS)
-  if context.index >= 4, context.props.selectedPage != context.tabData.key {
+  if context.props.filteredItems.count > 5, context.index >= 4, context.props.selectedPage != context.tabData.key {
     context.onSelect(context.tabData.key)
   }
 #endif
```

## Behavior

- **>5 tabs (More menu):** unchanged — the appear-sync still runs for overflowed tabs.
- **≤5 tabs:** the spurious force-select no longer fires, so the 5th tab stops hijacking the selection on re-appear.

## Testing

Verified on a 5-tab app (`@bottom-tabs/react-navigation` + expo-router) with a real device build: returning from a `tabBarHidden` screen no longer jumps to the 5th tab; normal tab switching and the >5-tab More flow are unaffected.

### DIFF
--- a/.changeset/bright-taxes-cheat.md
+++ b/.changeset/bright-taxes-cheat.md
@@ -1,0 +1,5 @@
+---
+'react-native-bottom-tabs': patch
+---
+
+Fix phantom tab switch on iOS when scene re-appears
--- a/packages/react-native-bottom-tabs/ios/TabAppearModifier.swift
+++ b/packages/react-native-bottom-tabs/ios/TabAppearModifier.swift
@@ -18,7 +18,12 @@ struct TabAppearModifier: ViewModifier {
       #endif
 
       #if os(iOS)
-        if context.index >= 4, context.props.selectedPage != context.tabData.key {
+        // Sync selection for tabs nested under the system "More" tab, which
+        // only exists when there are more than 5 visible tabs. Without the
+        // count guard the 5th tab (index 4) of a non-overflowing bar would
+        // force-select itself whenever its scene re-appears (e.g. when the tab
+        // bar is re-shown after `.hideTabBar`), hijacking the selection.
+        if context.props.filteredItems.count > 5, context.index >= 4, context.props.selectedPage != context.tabData.key {
           context.onSelect(context.tabData.key)
         }
       #endif


thiagobrez:
I tested this locally and it works nicely. Thanks @davidecallegaro! 🚀 

I will be merging soon

thiagobrez (approved):

thiagobrez:
Released in `1.3.0` 🎉

Issue #534, opened 2026-06-11
Badges not showing on certain pages on docs website


thiagobrez:
On the docs website, in some pages, badges like `<Badge text="iOS" type="info" /> <Badge text="experimental" type="danger"/>` are not rendering.

Correct example ("Usage with React Navigation"):

<img width="919" height="363" alt="Image" src="https://github.com/user-attachments/assets/46651642-d9de-4612-a603-f769314794f0" />

Incorrect example ("Standalone usage"):

<img width="909" height="372" alt="Image" src="https://github.com/user-attachments/assets/6365b314-ee71-468c-8af2-4df85a255294" />

Thecesar85:
Linked PR: #537\n\nOne-line summary: convert the standalone guide to MDX and import Badge so the inline badges render correctly.

thiagobrez:
Thanks @Thecesar85, PR looks good!

One suggestion though, keep writing your comments manually ☝🏻 that bad formatting and "summary" makes it look cheap

Fixed in #537

Thecesar85:
Hi @thiagobrez! Thank you again for the feedback and for merging PR #537.

I am still very new to this side of open source, so I wanted to ask something simple: was this issue connected to any bounty or reward process, or was it just a normal contribution?

If there is any claim or payout step involved, I would be grateful if you could point me in the right direction. And if not, that is totally fine too. I am mainly trying to learn how these situations are usually handled.

Thanks again for the review and the merge.

thiagobrez:
@Thecesar85 unfortunately no bounties involved. Thanks for the contribution!

Thecesar85:
@thiagobrez thank you very much for clarifying this.

Either way, I appreciate the guidance. Even if there is no payout path here, this was still useful experience for me and helped me better understand the project workflow.

PR #537, opened 2026-06-11
fix(#534): render badges in standalone docs guide

Thecesar85:
## Summary
- rename the standalone usage guide from Markdown to MDX so inline JSX components are parsed
- import `Badge` from `@theme` at the top of the guide

## Why
The page already uses inline `<Badge />` components, but plain Markdown does not evaluate them. Converting the guide to MDX makes those badges render the same way they do in the other docs pages.

## Validation
- `node .yarn/releases/yarn-4.6.0.cjs workspace react-native-bottom-tabs-docs build`


thiagobrez (approved):
Awesome, thank you!

PR #538, opened 2026-06-11
fix: avoid deprecated object form of CommonActions.navigate on tab press

rayabelcode:
Since `@react-navigation/routers` 7.6.0 (pulled in by `@react-navigation/native` 7.3.x), passing an object to `CommonActions.navigate` is deprecated and logs a warning. Because `NativeBottomTabView` dispatches the object form on tab selection, every tab press in an app on React Navigation 7.3+ now prints:

```
WARN  Passing an object as the argument to 'navigate' is deprecated. Use 'navigate(name, params, options)' instead.
```

This swaps it for the supported `navigate(name, params)` signature. It's the same pattern `@react-navigation/bottom-tabs` uses for its own tab press handling (`BottomTabBar.tsx`):

```js
navigation.dispatch({
  ...CommonActions.navigate(route.name, route.params),
  target: state.key,
});
```

Behavior is unchanged: the action is still scoped to the tab navigator via `target: state.key`, and the route is resolved by name within that navigator exactly as before. The only difference is the route's `key` is no longer spread into the action payload, which the v7 navigate action doesn't use for matching anyway.

Repro: any app using `createNativeBottomTabNavigator` with `@react-navigation/native >= 7.3.0`, press any unfocused tab, watch the console.

### DIFF
--- a/.changeset/fifty-taxes-go.md
+++ b/.changeset/fifty-taxes-go.md
@@ -1,0 +1,5 @@
+---
+'@bottom-tabs/react-navigation': patch
+---
+
+Fix deprecated object argument on CommonActions.navigate
--- a/apps/example/package.json
+++ b/apps/example/package.json
@@ -14,7 +14,7 @@
     "@bottom-tabs/react-navigation": "*",
     "@react-navigation/bottom-tabs": "^7.15.9",
     "@react-navigation/devtools": "^7.0.44",
-    "@react-navigation/native": "^7.2.2",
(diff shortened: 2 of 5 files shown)


thiagobrez (approved):

thiagobrez:
Thanks @rayabelcode !!

thiagobrez:
Released in `1.3.1` 🎉

rayabelcode:
Thanks @thiagobrez!

PR #540, opened 2026-06-16
Fix/expo template install error

thiagobrez:
## PR Description

Fix error on install of `expo-template`.


thiagobrez:
Released in `1.3.1` 🎉

PR #543, opened 2026-06-18
feat(iOS): use focused tab icons for selected state

thiagobrez:
## Summary

Adds focused icon metadata to the native tab view so iOS can use the focused tab icon for the selected/hover state instead of continuing to display the unfocused image.

https://github.com/user-attachments/assets/63872895-fcf6-485a-9f8e-1d6be9ec3b4b


thiagobrez:
Released on `1.4.0` 🎉

PR #549, opened 2026-07-01
fix(android): keep BottomNavigationView pinned under the keyboard in edge-to-edge

### LINKED ISSUE #357: [Android] Tab bar sticks to keyboard instead of being hidden behind it
### Before submitting a new issue

- [x] I tested using the latest version of the library, as the bug might be already fixed.
- [x] I tested using a [supported version](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md) of react native.
- [x] I checked for possible duplicate issues, with possible answers.

### Bug summary

The tab bar sticks to keyboard on Android on Chat layout. It happens even on example project in this repo:

Image

The problem is that textinput is also hidden behind tab bar.

### Library version

0.9.1

### Environment info

```shell
System:
  OS: macOS 15.1
  CPU: (14) arm64 Apple M4 Pro
  Memory: 140.16 MB / 48.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 22.14.0
    path: ~/.nvm/versions/node/v22.14.0/bin/node
  Yarn: Not Found
  npm:
    version: 10.9.2
    path: ~/.nvm/versions/node/v22.14.0/bin/npm
  Watchman: Not Found
Managers:
  CocoaPods:
    version: 1.16.2
    path: /Users/rosskhanas/.rbenv/shims/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 24.2
      - iOS 18.2
      - macOS 15.2
      - tvOS 18.2
      - visionOS 2.2
      - watchOS 11.2
  Android SDK: Not Found
IDEs:
  Android Studio: 2024.3 AI-243.24978.46.2431.13208083
  Xcode:
    version: 16.2/16C5032a
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.14
    path: /usr/bin/javac
  Ruby:
    version: 3.1.6
    path: /Users/rosskhanas/.rbenv/shims/ruby
npmPackages:
  "@react-native-community/cli": Not Found
  react: Not Found
  react-native:
    installed: 0.77.1
    wanted: ^0.77.0
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: false
  newArchEnabled: true
iOS:
  hermesEnabled: false
  newArchEnabled: false
```

### Steps to reproduce

1. Run example project on Android, I use Pixel 9 emulator
2. Click on a "Four Tabs" button
3. Open Chats
4. Focus text input

### Reproducible sample code

```js
https://github.com/callstackincubator/react-native-bottom-tabs/blob/main/apps/example/src/Screens/Chat.tsx
```

rayabelcode:
## What

Under edge-to-edge on Android, focusing a text input pushes the native tab bar up on top of the keyboard instead of letting the keyboard overlay it (iOS behavior). Fixes #357.

## Why

The library delegates inset handling to Material. `BottomNavigationView` installs a default inset listener in its constructor that adds `getSystemWindowInsetBottom()` to its bottom padding, and that value includes the IME inset. With `decorFitsSystemWindows=false` the window is not resized, so the IME inset propagates down to the bar and Material pads it by the keyboard height every frame (material-components/material-components-android#493). `windowSoftInputMode="adjustPan"` does not help when `react-native-keyboard-controller` is present, since it forces `adjustResize` at runtime.

## How

Give `ExtendedBottomNavigationView` its own `OnApplyWindowInsetsListener` that applies system bars + display cutout insets only, excluding `Type.ime()`, and returns the insets unconsumed so sibling views (the tab-screen content that hosts inputs) still receive the IME inset. Gesture-nav clearance is preserved because system-bar and cutout insets are still applied as padding.

This is a no-op outside edge-to-edge: with `decorFitsSystemWindows=true` the decor consumes the system insets before they reach the bar, so the computed insets are zero and padding stays at its base, and the keyboard resizes the window rather than dispatching an IME inset. Behavior only changes for the edge-to-edge case that is currently broken.

## Test plan

- Edge-to-edge app, tab-root screen with a top input: focus the input, confirm the tab bar stays pinned and the keyboard overlays it.
- Form screen with a bottom input inside a tab: confirm keyboard avoidance still works.
- Non-edge-to-edge build: confirm no change (bar still clears the nav bar, keyboard resizes as before).
- Landscape / gesture-nav and 3-button nav: confirm the bar still clears the system nav bar.

### DIFF
--- a/packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabView.kt
+++ b/packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabView.kt
@@ -21,6 +21,8 @@ import android.widget.FrameLayout
 import android.widget.LinearLayout
 import android.widget.TextView
 import androidx.core.view.MenuItemCompat
+import androidx.core.view.ViewCompat
+import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.forEachIndexed
 import coil3.ImageLoader
 import coil3.asDrawable
@@ -40,6 +42,29 @@ import com.google.android.material.navigation.NavigationBarView.LABEL_VISIBILITY
 import com.google.android.material.transition.platform.MaterialFadeThrough
 
 class ExtendedBottomNavigationView(context: Context) : BottomNavigationView(context) {
+  init {
+    // Material's BottomNavigationView adds the bottom system-window inset (which includes
+    // the IME inset under edge-to-edge) to its own padding, so the bar rises above the
+    // keyboard. Re-listen with system bars + display cutout only, excluding the IME type,
+    // so the bar stays pinned and the keyboard overlays it, matching iOS.
+    val baseLeft = paddingLeft
+    val baseTop = paddingTop
+    val baseRight = paddingRight
+    val baseBottom = paddingBottom
+    ViewCompat.setOnApplyWindowInsetsListener(this) { view, insets ->
+      val bars = insets.getInsets(
+        WindowInsetsCompat.Type.systemBars() or WindowInsetsCompat.Type.displayCutout(),
+      )
+      view.setPadding(
+        baseLeft + bars.left,
+        baseTop,
+        baseRight + bars.right,
+        baseBottom + bars.bottom,
+      )
+      insets
+    }
+  }
+
   override fun getMaxItemCount(): Int {
     return 100
   }


HaraldHenriksson:
We independently root-caused this exact bug in a production React Native app and arrived at the same fix as this PR before finding it — confirming the approach from a second angle.

**New data worth having for the merge decision:** this now reproduces even in apps that were previously shielded. Our app wraps everything in `react-native-keyboard-controller`'s `KeyboardProvider`, whose root-level IME-inset consumption used to prevent the insets from ever reaching `BottomNavigationView`. On the newest OS builds, consumed insets are dispatched to descendants anyway, so the Material bar keyboard-avoids regardless:

| Device | OS | Tab bar above keyboard? |
|---|---|---|
| Pixel 10 Pro | Android 17 | 🔴 yes |
| Stock emulator | API 37.1 (Android 17) | 🔴 yes |
| Galaxy S26 Ultra | Android 16, newest One UI 8.x | 🔴 yes |
| Galaxy S22 Ultra | Android 16, One UI 8.0 | 🟢 no |
| Emulator | API 35 | 🟢 no |

Same APK everywhere — the trigger is purely the OS build, so this is switching on device-by-device as Android 17 / One UI updates roll out. All input fields, every screen with the native bar.

**Verification of this PR's fix:** we shipped the identical change via patch-package and confirmed on-device — Android 17 emulator, S26 Ultra and Pixel 10 Pro fixed (keyboard overlays the bar again), and no visual change at rest on the previously-working devices (system-bar/cutout padding preserved). Would love to see this merged + released so we can drop the patch.

matinzd:
I tested this diff on Pixel 9 and it's working for me. Can someone look into this?

markrickert:
I applied this PR changes as a patch to my local project and it fixed my keyboard issues with the tab bar floating above the keyboard!

rayabelcode:
> I applied this PR changes as a patch to my local project and it fixed my keyboard issues with the tab bar floating above the keyboard!

Glad it helped! It's still working locally for me as well, hopefully it gets pulled into main soon.

PR #558, opened 2026-07-23
fix: report content size to JS when tab bar visibility changes on Android

### LINKED ISSUE #557: [Android] Content doesn't resize when tab bar visibility changes (tabBarHidden)
### Before submitting a new issue

- [x] I tested using the latest version of the library, as the bug might be already fixed.
- [x] I tested using a supported version of react native.
- [x] I checked for possible duplicate issues, with possible answers.

### Bug summary

On Android, when `tabBarHidden` is toggled to `true`, the native bottom navigation view is set to `GONE` and its space is reclaimed at the native level, but the React content does not grow to fill it. The screen keeps its old height, leaving a blank strip at the bottom exactly the height of the tab bar. Bottom-anchored elements (`position: 'absolute', bottom: 0`) end up floating ~80dp above the actual bottom of the screen. Toggling back to visible has the same problem in reverse until something else triggers a relayout.

I tracked it down to the layout listener in `RCTTabView.kt`. The size-change guard compares the bounds of the whole container (`ReactBottomNavigationView`), but when only the tab bar visibility changes, the container keeps the exact same size — only the inner `layoutHolder` grows or shrinks:

```kotlin
post {
  addOnLayoutChangeListener { _, left, top, right, bottom, _, _, _, _ ->
    val newWidth = right - left      // container size — unchanged when tab bar is hidden
    val newHeight = bottom - top

    onTabBarMeasuredListener?.invoke(...)

    if (newWidth != lastReportedSize?.width || newHeight != lastReportedSize?.height) {
      // never reached in this case, so onNativeLayoutListener doesn't fire
      onNativeLayoutListener?.invoke(dpWidth, dpHeight)
      lastReportedSize = Size(newWidth, newHeight)
    }
  }
}
```

Two problems compound here:

1. The guard compares the container size, while the value actually reported to JS is the `layoutHolder` size. When the tab bar is hidden, the container stays the same but `layoutHolder` changes, so the guard filters the update out.
2. The listener is attached to the container, and `OnLayoutChangeListener` only fires when that view's own bounds change — which they don't in this scenario. So even the callback itself may not run.

As a result `onNativeLayout` never reaches the JS side, `setMeasuredDimensions` keeps the stale height, and the children rendered by `TabView` stay at the old size.

Verified on an emulator with `uiautomator dump`: after hiding the tab bar, the native `layoutHolder` grows to full screen height (2400px on my device), while the React subtree inside it stays at the previous 2126px.

iOS is not affected.

### Library version

1.4.0

### Environment info

```shell
System:
  OS: macOS 26.5.2
  CPU: (18) arm64 Apple M5 Pro
Binaries:
  Node: 24.13.0
  Yarn: 3.6.4
  npm: 11.6.2
npmPackages:
  react: 19.2.3
  react-native: 0.86.0
Android:
  hermesEnabled: true
  newArchEnabled: true
iOS:
  hermesEnabled: true
  newArchEnabled: true
```

### Steps to reproduce

1. Android app using `createNativeBottomTabNavigator` (New Architecture).
2. Render a screen with a bottom-anchored view, e.g. an action bar with `position: 'absolute', bottom: 0`.
3. Toggle `tabBarHidden` to `true` (typical case: entering a selection/edit mode).
4. The tab bar disappears, but the bottom-anchored view stays where the tab bar's top edge used to be, with an empty strip below it.

Tested on an Android 16 emulator (arm64), RN 0.86, and reproduced consistently.

### Reproducible sample code

```tsx
const [hidden, setHidden] = useState(false);

<TabView tabBarHidden={hidden} ...>
  {/* screen content */}
  <View style={{ position: 'absolute', bottom: 0, left: 0, right: 0, height: 56 }}>
    <Button title="toggle" onPress={() => setHidden(v => !v)} />
  </View>
</TabView>
```

I have a fix working locally (report the `layoutHolder` size instead, and also listen to `layoutHolder` layout changes so the visibility-toggle case is caught). Will open a PR.

JJalit:
## PR Description

Fixes #557

Bug fix. On Android, toggling `tabBarHidden` didn't resize the React content. The layout listener's size-change guard compares the bounds of the whole `ReactBottomNavigationView` container, but hiding/showing the tab bar doesn't change the container bounds — only the inner `layoutHolder` grows or shrinks. On top of that, the listener is attached to the container, so `OnLayoutChangeListener` doesn't even fire in that case. As a result `onNativeLayoutListener` never reported the new size, `setMeasuredDimensions` on the JS side kept the stale height, and the content stayed short with a blank strip where the tab bar used to be.

Changes:

- Compare and report the `layoutHolder` size (which is what was already being sent to JS anyway) instead of the container size, extracted into `reportLayoutHolderSizeIfChanged()`.
- Also observe `layoutHolder` layout changes, so the visibility-toggle case (container bounds unchanged) is caught.

## How to test?

1. In an Android app using `createNativeBottomTabNavigator`, render a screen with a bottom-anchored view (`position: 'absolute', bottom: 0`).
2. Toggle `tabBarHidden` to `true`.
3. Before this change: the bottom-anchored view stays at the tab bar's old top edge, with an empty strip below it. After: the content fills the reclaimed space and the view sits at the actual bottom of the screen.
4. Toggle back to `false` — the content shrinks back above the tab bar.

Verified on an Android 16 emulator (RN 0.86, New Architecture) with `uiautomator dump`: previously the React subtree stayed at 2126px while the native `layoutHolder` grew to 2400px after hiding the tab bar; with this change both match, in both directions.

## Screenshots

N/A — measurement/layout change, behavior described above.

### DIFF
--- a/.changeset/tall-donkeys-resize.md
+++ b/.changeset/tall-donkeys-resize.md
@@ -1,0 +1,5 @@
+---
+'react-native-bottom-tabs': patch
+---
+
+Fix Android content not resizing when the tab bar visibility changes (`tabBarHidden`). The layout listener compared the container bounds, which stay the same when only the tab bar is hidden or shown, so the new `layoutHolder` size was never reported to JS and the content kept its stale height.
--- a/packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabView.kt
+++ b/packages/react-native-bottom-tabs/android/src/main/java/com/rcttabview/RCTTabView.kt
@@ -100,22 +100,33 @@ class ReactBottomNavigationView(context: Context) : LinearLayout(context) {
     uiModeConfiguration = resources.configuration.uiMode
 
     post {
-      addOnLayoutChangeListener { _, left, top, right, bottom,
+      addOnLayoutChangeListener { _, _, _, _, _,
                                   _, _, _, _ ->
-        val newWidth = right - left
-        val newHeight = bottom - top
-
         // Notify about tab bar height.
         onTabBarMeasuredListener?.invoke(Utils.convertPixelsToDp(context, bottomNavigation.height).toInt())
 
-        if (newWidth != lastReportedSize?.width || newHeight != lastReportedSize?.height) {
-          val dpWidth = Utils.convertPixelsToDp(context, layoutHolder.width)
-          val dpHeight = Utils.convertPixelsToDp(context, layoutHolder.height)
-
-          onNativeLayoutListener?.invoke(dpWidth, dpHeight)
-          lastReportedSize = Size(newWidth, newHeight)
-        }
+        reportLayoutHolderSizeIfChanged()
       }
+      // When only the tab bar visibility changes (tabBarHidden), the container keeps
+      // its bounds so the listener above never fires — only layoutHolder grows or
+      // shrinks. Observe layoutHolder itself so the new content size reaches JS.
+      layoutHolder.addOnLayoutChangeListener { _, _, _, _, _,
+                                               _, _, _, _ ->
+        reportLayoutHolderSizeIfChanged()
+      }
+    }
+  }
+
+  private fun reportLayoutHolderSizeIfChanged() {
+    val newWidth = layoutHolder.width
+    val newHeight = layoutHolder.height
+
+    if (newWidth != lastReportedSize?.width || newHeight != lastReportedSize?.height) {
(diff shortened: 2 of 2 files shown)


markrickert:
👍 
Thank you for this PR! I can confirm that this fixed my issue on Android when swapping the `tabBarHidden` value back and forth!

I've applied a patch to my project till this fix ships!

PR #581, opened 2026-09-11
feat:migrate to standard navigation shared adapter

oleksandrzavarzin-callstack:
## What

Expo SDK 56 dropped Expo Router's dependency on `@react-navigation/*` in favour of a vendored fork. `@bottom-tabs/react-navigation` imports the real thing, so on SDK 56+ two copies of React Navigation load and their contexts stop lining up.

This moves the tab view into a new framework-agnostic package, `@bottom-tabs/standard-navigation`, built on the `standard-navigation` contract. One implementation, two adapters:

| Expo SDK | Package                            | Integration                              |
| -------- | ---------------------------------- | ---------------------------------------- |
| 56+      | `@bottom-tabs/standard-navigation` | `unstable_createStandardRouterNavigator` |
| 52 to 55 | `@bottom-tabs/react-navigation`    | `withLayoutContext`                      |

**`@bottom-tabs/react-navigation`'s public API is unchanged** - same exports, same types, same theme-derived tint defaults. It becomes a thin adapter over the shared view.

## Notable

- Dropped the `color` dependency for React Native's `processColor`. `color` is ESM-only, so every consumer who writes tests had to patch `transformIgnorePatterns` before Jest would run at all.
- The peer on `@react-navigation/native` stays at `>=7`. An earlier revision raised it to `>=7.3.0`, which broke `npm install` outright for SDK 52-55 apps locked below that - and nothing on that path needs 7.3.
- `NativeBottomTabsContent` carries its event map and navigator props on phantom type-only properties. Without them an integrator loses every navigator prop at the call site: `tabBarActiveTintColor` and friends become type errors. 
- `unstable_createStandardRouterNavigator` is marked unstable by Expo and may change between minors. We deliberately don't call it ourselves, so a break stays in app code.
- On the Expo Router path a custom `tabBar` receives `{ state, descriptors, actions, emitter }` rather than a `navigation` object, so `@react-navigation/bottom-tabs`' `BottomTabBar` can't be used there. The React Navigation path is unaffected.

## How to test

```sh
yarn && yarn build
yarn test        # 96 tests across 5 workspaces
yarn lint && yarn typecheck
```

`apps/expo-router-test` runs one assertion suite against **both** adapters under a real Expo Router tree, with a console spy that fails on any warning or error.

Verified on an app scaffolded from `create-expo-app --template default@sdk-57` (Expo 57, React Native 0.86.3, React 19.2.3):

- **iOS 26** and **Android 17** - builds and bundles, Liquid Glass and Material 3 tab bars render with tint and badge, tapping routes between screens. No JS warnings or errors.
- `@react-navigation/*` is absent from `node_modules` entirely, so there's no second copy to line up - everything resolves through Expo Router's vendored fork.

The SDK 52-55 path was checked separately: the documented `withLayoutContext` recipe typechecks clean against a real `expo-router@55` install, at both 7.1.33 and 7.3.0.

## Screenshots

| iOS 26 | Android 17 |
| --- | --- |
| image | image |

### DIFF
--- a/.changeset/standard-navigation-shared-adapter.md
+++ b/.changeset/standard-navigation-shared-adapter.md
@@ -3,7 +3,9 @@
 '@bottom-tabs/react-navigation': minor
 ---
 
-Move the native tabs implementation into a new framework-agnostic `@bottom-tabs/navigation` package built on the `standard-navigation` contract. This lets Expo Router SDK 56+ apps drive the same navigator through `unstable_integrateWithRouter` without loading a second copy of React Navigation.
+Move the native tabs implementation into a new framework-agnostic `@bottom-tabs/navigation` package built on the `standard-navigation` contract. This lets Expo Router SDK 56+ apps drive the same navigator through `unstable_createStandardRouterNavigator` without loading a second copy of React Navigation.
+
+`NativeBottomTabsContent` carries its event map and navigator props on phantom properties, so an integrator can type `tabBarActiveTintColor`, `tabBar` and the rest. Neither is inferable from the contract on its own - the event map appears only as an argument to `emitter.emit`, and the navigator props only inside an `Omit<…>`.
 
 `createNativeBottomTabNavigator` renders through the shared navigator, and keeps building its own state with `useNavigationBuilder`, so `state.key` and the per-route `navigation` and `route` objects a custom `tabBar` reads stay intact.
 
--- a/.changeset/standard-navigation-shared-adapter.md
+++ b/.changeset/standard-navigation-shared-adapter.md
@@ -3,4 +3,8 @@
 '@bottom-tabs/react-navigation': minor
 ---
 
-Move the native tabs implementation into a new framework-agnostic `@bottom-tabs/navigation` package built on the `standard-navigation` contract, and delegate the React Navigation entry point to it. This lets Expo Router SDK 56+ apps drive the same navigator through `unstable_integrateWithRouter` without loading a second copy of React Navigation
+Move the native tabs implementation into a new framework-agnostic `@bottom-tabs/navigation` package built on the `standard-navigation` contract. This lets Expo Router SDK 56+ apps drive the same navigator through `unstable_integrateWithRouter` without loading a second copy of React Navigation.
+
+`createNativeBottomTabNavigator` renders through the shared navigator, and keeps building its own state with `useNavigationBuilder`, so `state.key` and the per-route `navigation` and `route` objects a custom `tabBar` reads stay intact.
+
+`@bottom-tabs/react-navigation` now requires `@react-navigation/native` 7.3.0 or newer, which is the version that introduced `createStandardNavigationFactories` used by the documented shared-navigator recipe
--- a/.changeset/standard-navigation-shared-adapter.md
+++ b/.changeset/standard-navigation-shared-adapter.md
@@ -1,0 +1,6 @@
+---
+'@bottom-tabs/navigation': minor
+'@bottom-tabs/react-navigation': minor
+---
+
+Move the native tabs implementation into a new framework-agnostic `@bottom-tabs/navigation` package built on the `standard-navigation` contract, and delegate the React Navigation entry point to it. This lets Expo Router SDK 56+ apps drive the same navigator through `unstable_integrateWithRouter` without loading a second copy of React Navigation
--- a/apps/expo-router-test/src/__tests__/adapters.test.tsx
+++ b/apps/expo-router-test/src/__tests__/adapters.test.tsx
@@ -1,0 +1,116 @@
+import { NavigationContainer } from '@react-navigation/native';
+import { act, render } from '@testing-library/react-native';
+import { renderRouter } from 'expo-router/testing-library';
+import { Text } from 'react-native';
+
+import { ExpoRouterTabs, ReactNavigationTabs } from '../tabs';
+
+let tabViewProps: Record<string, any>;
+
+jest.mock('react-native-bottom-tabs', () => ({
+  __esModule: true,
+  default: (props: Record<string, any>) => {
+    tabViewProps = props;
+    return null;
+  },
+}));
+
+const OPTIONS = {
+  index: { title: 'Home', tabBarBadge: '3' },
+  explore: { title: 'Explore' },
+};
+
+const adapters = [
+  {
+    name: 'expo-router',
+    render: async () => {
+      await renderRouter(
+        {
+          _layout: () => (
+            <ExpoRouterTabs>
+              <ExpoRouterTabs.Screen name="index" options={OPTIONS.index} />
+              <ExpoRouterTabs.Screen name="explore" options={OPTIONS.explore} />
+            </ExpoRouterTabs>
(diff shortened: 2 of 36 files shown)


.changeset/standard-navigation-shared-adapter.md
```diff
@@ -0,0 +1,12 @@
+---
+'@bottom-tabs/navigation': minor
```
thiagobrez: Let's call this new package `@bottom-tabs/standard-navigation` to keep consistency with React Navigation

.changeset/standard-navigation-shared-adapter.md
```diff
@@ -0,0 +1,12 @@
...
+'@bottom-tabs/react-navigation': minor
+---
+
+Move the native tabs implementation into a new framework-agnostic `@bottom-tabs/navigation` package built on the `standard-navigation` contract. This lets Expo Router SDK 56+ apps drive the same navigator through `unstable_createStandardRouterNavigator` without loading a second copy of React Navigation.
+
+`NativeBottomTabsContent` carries its event map and navigator props on phantom properties, so an integrator can type `tabBarActiveTintColor`, `tabBar` and the rest. Neither is inferable from the contract on its own - the event map appears only as an argument to `emitter.emit`, and the navigator props only inside an `Omit<…>`.
+
+`createNativeBottomTabNavigator` renders through the shared navigator, and keeps building its own state with `useNavigationBuilder`, so `state.key` and the per-route `navigation` and `route` objects a custom `tabBar` reads stay intact.
+
+`@bottom-tabs/react-navigation` now requires `@react-navigation/native` 7.3.0 or newer, which is the version that introduced `createStandardNavigationFactories` used by the documented shared-navigator recipe
```
thiagobrez: All of this is unnecessary for the changeset

docs/docs/docs/getting-started/quick-start.mdx
```diff
@@ -15,10 +15,14 @@ npx create-expo-app@latest NativeTabs --template @bottom-tabs/expo-template
 
 <PackageManagerTabs command="install react-native-bottom-tabs" />
 
-If you are going to use [React Navigation / Expo Router Integration](/docs/guides/usage-with-react-navigation) make sure to install `@bottom-tabs/react-navigation`.
+If you are going to use the [React Navigation integration](/docs/guides/usage-with-react-navigation), also install `@bottom-tabs/react-navigation`.
```
thiagobrez, line 18: Shouldn't it be here React Navigation or Expo SDK < 56? Or I'm missing something?

apps/expo-router-test/src/__tests__/adapters.test.tsx
```diff
@@ -0,0 +1,116 @@
+import { NavigationContainer } from '@react-navigation/native';
```
thiagobrez, line 1: Not related to this file specifically, but the example app should be named `example-expo-router` to follow the convention of the other example apps

docs/docs/docs/guides/usage-with-expo-router.mdx
```diff
@@ -2,61 +2,63 @@ import { PackageManagerTabs } from '@theme';
...
+Which package you need depends on your Expo SDK version, because Expo Router stopped depending on React Navigation starting from SDK 56.
 
-<PackageManagerTabs command="install @bottom-tabs/react-navigation" />
+| Expo SDK     | Package                            | Integration                              |
+| ------------ | ---------------------------------- | ---------------------------------------- |
+| 56 and newer | `@bottom-tabs/standard-navigation` | `unstable_createStandardRouterNavigator` |
+| 55 and older | `@bottom-tabs/react-navigation`    | `withLayoutContext`                      |
 
-Next, create a custom layout adapter for the native bottom tabs using `withLayoutContext` from Expo Router. Put it in `components/bottom-tabs.tsx` and export the navigator, so your layouts can import it:
+## Expo SDK 56 and newer
```
thiagobrez, line 12: Hard to see in the diff, but did we loose the section for integrating with SDK < 56?

It's important to have it since we still support it with the old integration

docs/docs/docs/getting-started/quick-start.mdx
oleksandrzavarzin-callstack: Yep, that's true. 
I've added "Expo SDK 55+" note to it.

apps/expo-router-test/src/__tests__/adapters.test.tsx
oleksandrzavarzin-callstack: It's a test harness rather than the example Expo router app (for that option, we will have `expo-template`, after it's updated). It just checks whether our two adapters, driven by a real `Expo Router` and a real `React Navigation`, pass the same correct props to the tab view.

docs/docs/docs/guides/usage-with-expo-router.mdx
oleksandrzavarzin-callstack: Here is that section:
https://github.com/callstack/react-native-bottom-tabs/blob/a65309c00a260e49918780d008d2db4b3361cfa3/docs/docs/docs/guides/usage-with-expo-router.mdx#L85-L124

docs/docs/docs/guides/usage-with-expo-router.mdx
```diff
@@ -66,7 +70,59 @@ export default function TabLayout() {
...
-For props and more information, see the [React Navigation integration guide](/docs/guides/usage-with-react-navigation), which documents every accepted `tabBarIcon` source.
+> [!NOTE] `unstable_createStandardRouterNavigator` is marked unstable by Expo and may change between minor releases.
+
+### Tint colors
+
+There is no theme in the Expo Router integration, so the tab bar falls back to platform defaults. Pass `tabBarActiveTintColor` and `tabBarInactiveTintColor` on the navigator to set them yourself.
+
+### Custom tab bar
+
+The `tabBar` function receives `state`, `descriptors`, `actions` and `emitter` - the vocabulary of the navigation contract rather than a React Navigation `navigation` object. If your tab bar needs the router, call `useNavigation()` from `expo-router` inside it.
```
thiagobrez: Remove AI text

```suggestion
The `tabBar` function receives `state`, `descriptors`, `actions` and `emitter`. If your tab bar needs the router, call `useNavigation()` from `expo-router` inside it.
```

docs/docs/docs/guides/usage-with-expo-router.mdx
```diff
@@ -66,7 +70,59 @@ export default function TabLayout() {
...
+
+`BottomTabBar` from `@react-navigation/bottom-tabs` cannot be used here, since it expects a React Navigation `navigation` object.
```
thiagobrez: Remove

oleksandrzavarzin-callstack: Removed the line

docs/docs/docs/guides/usage-with-expo-router.mdx
oleksandrzavarzin-callstack: Applied the suggested change with the removal of the line underneath it.

PR #585, opened 2026-09-19
New role `prominent` (SDK27+)

retyui:
## PR Description

Add a new role `prominent`

1. Only one tab can receive the `prominent` treatment
2. Available from SDK27+

## How to test?

well...

## Screenshots

Screenshot 2026-09-19 at 23 48 33


packages/react-native-bottom-tabs/ios/TabViewProps.swift
```diff
@@ -30,12 +30,22 @@ internal enum MinimizeBehavior: String {
...
       return .search
+    case .prominent:
+      #if compiler(>=6.4)
+        if #available(iOS 27, macOS 27, visionOS 27, tvOS 27, *) {
+          return .prominent
+        }
+      #endif
+      // `.prominent` role was introduced in iOS 27 SDK; ignore it on older OS versions
+      // instead of crashing or silently mapping to the wrong role.
+      return nil
```
retyui, line 48: I'm not sure about fallback value, maybe it's better to return `search` for old sdk versions

retyui:
@thiagobrez could you please review this PR?

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
```diff
@@ -38,7 +38,7 @@ struct NewTabView: AnyTabView {
               onSelect: onSelect
             )
 
-            Tab(value: tabData.key, role: tabData.role?.convert()) {
+            Tab(value: tabData.key, role: tabData.role.flatMap { $0.convert() }) {
```
troZee: q: Why is it needed? It seems to be the same change

packages/react-native-bottom-tabs/ios/TabViewProps.swift
troZee: IMO we should fall back to the old approach if possible

packages/react-native-bottom-tabs/ios/TabView/NewTabView.swift
retyui: @troZee `convert()` itself returns an optional value, this result becomes a nested optional `TabRole??` (or `Optional<Optional<TabRole>>`) so , `flatMap` safely unwraps `tabData.role` if it exists

packages/react-native-bottom-tabs/ios/TabViewProps.swift
retyui: > back to the old approach if possible
I think it not possible
 
**case:** if a `prominent` will  be applied to the `+ add an item` tab
and fallback value for SDK 26 is what? `search`  ?

troZee: @retyui this is what I was thinking about but @thiagobrez must decide about it

thiagobrez: Let's keep `nil`. Fallback to `search` is not 1:1 and can be deceiving for users who understand the behavior 👍🏻

finn-pleo:
Nice, was just about to create a PR for this 👏  Potentially older versions could use `search` as a fallback for `prominent`, otherwise projects might need ` Number(Platform.Version) >= 27 ? "prominent" : "search"`. However I think it's also fine to handle that explicitly

thiagobrez:
@finn-pleo The behavior between `prominent` and `search` is different, so using `search` as a fallback can be deceiving for users who understand how it works.

I'm working through some bugs and this should be merged soon :)

